PR #45, opened 2015-06-02
Align

velo:
@rultor merge


hazendaz:
I think we would need to cherry pick bits and pieces.  Can you put up another PR based off your current?  I'll look and see what we can start bring in.  The structure changes would be the hardest but possible I can rule out commits that don't matter and try to get somewhat caught up.

--- Original Message ---

hazendaz:
@velo was intended.  He opened this.  Trying to directly merge code won't fly. We will need to cherry pick bits and pieces.  The commit line has various duplication and pretty large structure changes.

--- Original Message ---

PR #52, opened 2015-07-12
Removed site-maven-plugin as we are not using that

hazendaz:
We are using site:deploy for site distribution.  Removed
site-maven-plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -307,20 +307,6 @@
                     </dependencies>
                 </plugin>
                 <plugin>
-                    <groupId>com.github.github</groupId>
-                    <artifactId>site-maven-plugin</artifactId>
-                    <version>0.12</version>
-                    <dependencies>
-                        <!-- Site Maven Plugin is referencing incorrect version 
-                            so fixed here -->
-                        <dependency>
-                            <groupId>org.eclipse.mylyn.github</groupId>
-                            <artifactId>org.eclipse.egit.github.core</artifactId>
-                            <version>2.1.5</version>
-                        </dependency>
-                    </dependencies>
-                </plugin>
-                <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>2.18.1</version>


ctubbsii:
FWIW, in other projects, I've found it useful to use the site-maven-plugin, bound to the site-deploy phase. It might be easier than the current wagon plugins. Whatever works... it's an option, though.

hazendaz:
I use to use that.  Ran into issues though. The one we are using is more recent and works in more typical maven way.  Mvn site just builds local and mvn site:deploy will push it.  And it's bound to the mvn release as well.  I found the other would deploy every time I built the site.  The one I removed wasn't actually configured now.  It was left over.  Look through for the git based site wagon stuff.  Should get you out to that author.  Now I must admit this new way is slower but then again it's using ssh.


Sent by Outlook for Android



From: Christopher Tubbs

Sent: Thursday, July 16, 2:46 PM

Subject: Re: [formatter-maven-plugin] Removed site-maven-plugin as we are not using that (#52)

To: revelc/formatter-maven-plugin

Cc: Jeremy Landis



FWIW, in other projects, I've found it useful to use the site-maven-plugin, bound to the site-deploy phase. It might be easier than the current wagon plugins. Whatever works... it's an option, though.

ctubbsii:
> I found the other would deploy every time I built the site.

I got around that by configuring the plugin to bind to the site-deploy phase, and instead of executing `mvn site` to execute the [full site lifecycle][1], I would just execute `mvn site:site` or `mvn site:deploy`.

Your solution works, too. I'll have to look at the plugins you're using. I may want to switch to them instead.

[1]: https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference

hazendaz:
Smart solution!


Sent by Outlook for Android





On Fri, Jul 17, 2015 at 2:19 PM -0700, "Christopher Tubbs" <notifications@github.com> wrote:
> I found the other would deploy every time I built the site.

I got around that by configuring the plugin to bind to the site-deploy phase, and instead of executing `mvn site` to execute the [full site lifecycle][1], I would just execute `mvn site:site` or `mvn site:deploy`.

Your solution works, too. I'll have to look at the plugins you're using. I may want to switch to them instead.

[1]: https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference

---

PR #54, opened 2015-07-22
Velo merge

velo:
Wanna check if travis is no able to build this

PR #55, opened 2015-07-30
Velo 2

hazendaz:
Interesting, all original commits remain.  The stuff at the end was me organizing so it became easier to see what changed.  I'm having issues building this so please don't merge.  It's just something else to look at for the moment.


hazendaz:
Well that answers that.  I broke the eclipse download piece.  @velo do you see anything wrong on here?  I assume once I get past that, it will be back to the test cases I was having issues with or travis will say they are ok.

hazendaz:
Order on what I did for reference.

- Aligned the library packaging to match velo before merging that code
- Verified original still built correctly after some pom changes
- Merged with --no-ff for velo
- Fixed merge conflicts
- Then some other trivial things.  The issue at hand was present after resolving merge conflicts

Verified at this point the jars download for latest eclipse only have META-INF in them.  Remainder is missing.  So I'm losing the classes.  Could be a issue in the parent pom.

velo:
one thing that caught my attention, you are now the author of all files I created/change... 
which rendered blame/history useless
https://github.com/hazendaz/formatter-maven-plugin/blame/velo-2/m2e-configurator/net.revelc.code.formatter.tests/src/org/sonatype/m2e/buildhelper/tests/BuildhelperTest.java

velo:
@hazendaz your changes removed the maven-dependency-plugin...

Looks like your commits are just overwriting all files I create (like they were deleted and added again).
That makes way more difficult to find out what changed.

Anyway, I think, this is what is missing.
https://github.com/velo/maven-formatter-plugin/commit/7fecafa061098dfcd68688af38cde7aa8bb3ed71#diff-049bec822d734a54df89a0c1c10034e9R32

hazendaz:
I had that dependency plugin there but it doesn't do anything.  No execution on it same error.  Original commits are all there.    My goal was to see exactly what changed in plugin piece.  If I can get that and not recommit all your other modules that would be best.  Do you know how? Maybe a cherry pick so we just have a reference. Or maybe it's not important now we can look here to see the main difference then just merge velo merge as is tonight.  I can you save this as a branch in case we need to look at anything.


Sent by Outlook for Android





On Thu, Jul 30, 2015 at 12:14 AM -0700, "Marvin Froeder" <notifications@github.com> wrote:
@hazendaz your changes removed the maven-dependency-plugin...

Looks like your commits are just overwriting all files I create (like they were deleted and added again).
That makes way more difficult to find out what changed.

Anyway, I think, this is what is missing.
https://github.com/velo/maven-formatter-plugin/commit/7fecafa061098dfcd68688af38cde7aa8bb3ed71#diff-049bec822d734a54df89a0c1c10034e9R32

---

velo:
if you look at my PR, there are 3 hits on maven-dependency-plugin (well, only 2 are relevant, anyway)
https://github.com/revelc/formatter-maven-plugin/pull/54/files

hazendaz:
Closing this PR.  Will comment on original.

PR #57, opened 2015-07-31
[velo-merge] Merge velo code as new starting point

hazendaz:
Merging [this](https://github.com/velo/maven-formatter-plugin) as new starting point for project.  This ignores the commit history where forks diverged and accepts that @velo direction is the future of this project.


hazendaz:
Alternative to PR #54

velo:
doing it like this we loose all blame/history.... it prefer to have a ton of useless history to no history

PR #58, opened 2015-08-01
Velo merge - updates to velo-merge branch

hazendaz:


hazendaz:
How do we suppress the warnings out of proguard?  It's way too many for travis CI thus it fails.

hazendaz:
Top commit needs a double check.  Otherwise this now properly builds on my machine and issues with proguard which were problems in original are now fixed.  Along with a lot of minor things.  Please take a quick look at the top comit @velo and let me understand better what the test is actually testing.  I will be running some further tests on this tomorrow.  It should now solve me jar file config as well.

m2e-configurator/net.revelc.code.formatter/src/net/revelc/code/formatter/connector/internal/FormatterProjectConfigurator.java
```diff
@@ -106,20 +105,18 @@ public void configure(ProjectConfigurationRequest request,
 	private Xpp3Dom[] parseConfigurationFile(
 			ProjectConfigurationRequest request, IProgressMonitor monitor)
 			throws CoreException {
-		InputStream content = readConfigFile(Formatter.JAVA, request, monitor);
 		Xpp3Dom dom;
-		try {
+		try (InputStream content = readConfigFile(Formatter.JAVA, request, monitor)) {
```
velo: nice catch!

velo:
I added a comment on the commit.

All LGTM, so, merging!

hazendaz:
Hopefully I can get in and do some real testing tomorrow.  I think I have a glitch in reading in the Jar version of configuration.  Basically, I believe I caused an issue in what you had coded so will have to go back and look to see what other differenced lived in that area from original.  Hopefully get that all resolved soon.

PR #60, opened 2015-08-05
Velo merge

hazendaz:
A number of micro cleanups and fixing formatter to properly read from resource.  Only tested as much as a file at this point.  I will test actual jar location later.  This fixes a defect I introduced in the last PR.  There are still a number of things I'm seeing in formatter that are different from existing code base.  We still need to look more in this area.  This is just one step closer.

As a side note, it appears this requires both java / javascript setup to run.  Is that how it normally worked or due to my changes?

### DIFF
--- a/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -63,7 +63,6 @@ import net.revelc.code.formatter.java.JavaFormatter;
 import net.revelc.code.formatter.javascript.JavascriptFormatter;
 import net.revelc.code.formatter.model.ConfigReadException;
 import net.revelc.code.formatter.model.ConfigReader;
-import net.revelc.code.formatter.support.io.Resource;
 
 /**
  * A Maven plugin mojo to format Java source code using the Eclipse code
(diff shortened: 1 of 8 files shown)


maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
```diff
@@ -94,7 +94,7 @@
 	/**
 	 * Project's source directory as specified in the POM.
 	 */
-	@Parameter(defaultValue = "${project.build.sourceDirectory}", property = "sourceDirectory", required = true)
+	@Parameter(defaultValue = "${project.build.sourceDirectory}", readonly = true, required = true)
```
velo: this may not be true if you wanna the formatter to be run only once in a multi module project.

That is why I left this configurable.

hazendaz: Wouldn't skip do the same in child elements that do not need it?  I'm not sure in what use case a different source applies.  Or how that would make it run only once.  I guess if everything is considered source.  But I think that sort of violates maven convention.  I could be wrong though.  It's worth a second look.


Sent by Outlook for Android





On Tue, Aug 4, 2015 at 7:36 PM -0700, "Marvin Froeder" <notifications@github.com> wrote:
> @@ -94,7 +94,7 @@
>       /**
>        * Project's source directory as specified in the POM.
>        */
> -     @Parameter(defaultValue = "${project.build.sourceDirectory}", property = "sourceDirectory", required = true)
> +     @Parameter(defaultValue = "${project.build.sourceDirectory}", readonly = true, required = true)

this may not be true if you wanna the formatter to be run only once in a multi module project.

That is why I left this configurable.

---
https://github.com/revelc/formatter-maven-plugin/pull/60/files#r36265102

PR #64, opened 2015-08-07
Made eclipse config.xml files optional.

velo:
Revert "Rework resource usage"

This reverts commit bdd925b20a32e8c4c2631ac23eb30e482293d647.

### DIFF
--- a/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -63,6 +63,7 @@ import net.revelc.code.formatter.java.JavaFormatter;
 import net.revelc.code.formatter.javascript.JavascriptFormatter;
 import net.revelc.code.formatter.model.ConfigReadException;
 import net.revelc.code.formatter.model.ConfigReader;
+import net.revelc.code.formatter.support.io.Resource;
 
 /**
  * A Maven plugin mojo to format Java source code using the Eclipse code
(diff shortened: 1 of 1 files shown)


maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
```diff
@@ -583,7 +604,7 @@ private void createCodeFormatter() throws MojoExecutionException {
 		try {
 			this.resourceManager.addSearchPath(FileResourceLoader.ID,
 							this.basedir.getAbsolutePath());
-			configInput = this.resourceManager.getResourceAsInputStream(newConfigFile);
+			configInput = this.resourceManager.getResourceAsInputStream(this.configFile);
```
hazendaz: This was the bug I was mentioning.  The 'this' reference here is to the wrong field which would otherwise break usage for javascript anyway.  The this should be removed.  It should be using the 'configFile directly passed into the private method.

hazendaz:
It appears to me this was already done on master.  I found a few other minor related glitches in this for error logging.  Should we close this PR now?

velo:
No, this was not done... 

Since reverting was going to break other stuff I left it behind.

hazendaz:
Are you sure about that?  Because the code I wrote isn't on master, what was reverted is and it shows in the commit history.

e0ae79082f4b415ee0fb84844dd3540d07175f7e

hazendaz:
deleted my last message.   noted wrong commit.  I'm looking but certain the changes I had made are not on the master at all.

hazendaz:
Interesting.  velo-merge deleting deleted the PR.  It did put a revert branch out there with the changes so they are not lost.

PR #72, opened 2015-09-08
Pom updates, fix config error logging, and removed link from readme

hazendaz:


hazendaz:
Appveyor seems to take a long time.  This should be merged once that comes back.  It fixes a bug in logging and updates a few versions in pom.

Issue #13, opened 2015-03-14
Support using settings from maven-compiler-plugin


GoogleCodeExporter:
```
Like the maven-compiler-plugin, 
Use expression 'maven.compiler.source' for the compiler source/compliance 
params 
Use the expression 'maven.compiler.target' for the compiler target param
Continue to support the default version 1.5 for all params.

Incase the compiler plugin settings are specified in the plugin configuration 
(instead of the above expressions),
Parse the compiler plugin settings (similar to the maven-eclipse-plugin)
Get values for 'source' and 'target' (and even 'encoding' incase the 
'project.build.sourceEncoding' property is not used)

This helps to avoid needing to specify these values again in the formatter 
plugin configuration or move from compiler configuration to properties.
```

Original issue reported on code.google.com by `matt.blanchette@gmail.com` on 1 Mar 2011 at 8:24

PR #73, opened 2015-09-08
Made eclipse config.xml files optional.

hazendaz:
Revert "Rework resource usage"

This reverts commit bdd925b20a32e8c4c2631ac23eb30e482293d647.

For issue #13 

Originan PR was closed due to branch closing.  This has been rebased and the revert branch remains in case of any concerns.  This needs to be vetted so we don't lose original jar usage for configuration and gain what the original intent was.

### DIFF
--- a/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -63,6 +63,7 @@ import net.revelc.code.formatter.java.JavaFormatter;
 import net.revelc.code.formatter.javascript.JavascriptFormatter;
 import net.revelc.code.formatter.model.ConfigReadException;
 import net.revelc.code.formatter.model.ConfigReader;
+import net.revelc.code.formatter.support.io.Resource;
 
 /**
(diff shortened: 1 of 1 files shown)


hazendaz:
This currently won't apply successfully as it has an issue.  While I was able to get that fixed, it broke the jar version.  I've also got a few concerns that we are not really using the default ones we have set on the project within the project itself.  For one, PR is coming where I changed to use spaces instead of tabs per our original discussions and further using 120 line length.  Neither seem to apply at all.  Not sure why.  However, if I go back to 0.5.2 and use a wrapped jar with the same file contents, everything applies as expected.  If I leave it to use the ones under config it does not.  That leads me to believe that it's referencing the config we provide internally but location causes it to not be found and thus falls back.  Regardless something doesn't jive somewhere.  So at best I'm bringin in some trivial pull requests and maybe we can get this figured out here soon so we can push a release.  I've got a deep enough track record on 0.5.2 now that moving up shouldn't cause as many differences as I'm seeing especially when I'm using mars setup the same way and it is causing no differences between 0.5.2.  Feeling confused but this is where this stands at the moment... I didn't see anything obvious jumping out.

hazendaz:
Closing pull request as it's significantly out of date

PR #78, opened 2015-10-26
Validation fails when LineEnding.AUTO is used, even if actual line endings match

tisoft:


velo:
Can you provide a sample case as well?

hazendaz:
This is probably exactly the issue I'm facing.  I can't even get a build of the project to work as test cases fail due to line ending issues.  I'll test this one out locally tonight and provide feedback here.  If this fixes my build issue it will be merged.


Sent by Outlook for Android





On Mon, Oct 26, 2015 at 12:58 PM -0700, "Marvin Froeder" <notifications@github.com> wrote:
Can you provide a sample case as well?

---

hazendaz:
I feel the issue is that the enum comparison only applies to the enum values.  While they may match, AUTO isn't part of the determination and thus when AUTO is used it's going to be a mismatch.  Therefore this change is valid.

velo:
Yeah, but without any test is hard to prevent this from breaking again

hazendaz:
I tried to write one but no luck as it never returns auto but rather the actual type.  So it looks hard to test.


Sent by Outlook for Android





On Wed, Oct 28, 2015 at 9:10 PM -0700, "Marvin Froeder" <notifications@github.com> wrote:
Yeah, but without any test is hard to prevent this from breaking again

---

PR #80, opened 2015-10-29
Reduce restriction on builds, change line endings to auto, update plugins

hazendaz:
We need to validate only in CI builds - ie on travis ci not local builds as that ends up requiring developers to use eclipse to format which violates the spirit of the plugin.  If we don't enforce it there yet, we should.  For now, reducing the restriction locally.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -388,7 +388,7 @@
                 <executions>
                     <execution>
                         <goals>
-                            <goal>validate</goal>
+                            <goal>format</goal>
                         </goals>
                     </execution>
                 </executions>
--- a/pom.xml
+++ b/pom.xml
@@ -127,7 +127,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-clean-plugin</artifactId>
-                    <version>2.6.1</version>
+                    <version>3.0.0</version>
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -376,7 +376,7 @@
                         <exclude>**/src/test/resources/**</exclude>
                         <exclude>**/target/**</exclude>
                     </excludes>
-                    <lineEnding>LF</lineEnding>
+                    <lineEnding>AUTO</lineEnding>
                 </configuration>
                 <dependencies>
                     <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -121,8 +121,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-plugin-plugin</artifactId>
-                    <version>3.2</version>
-                    <!-- drop back from 3.3.3 as it breaks on lineEndings.java -->
+                    <version>3.4</version>
                 </plugin>
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -72,7 +72,7 @@
         </contributor>
     </contributors>
     <prerequisites>
-        <maven>3.1.1</maven>
+        <maven>3.2.5</maven>
     </prerequisites>
     <modules>
         <module>formatters</module>
@@ -110,7 +110,7 @@
     <properties>
         <maven.compiler.source>1.7</maven.compiler.source>
         <maven.compiler.target>1.7</maven.compiler.target>
-        <maven.min-version>3.1.1</maven.min-version>
+        <maven.min-version>3.2.5</maven.min-version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
--- a/maven-plugin/src/main/java/net/revelc/code/formatter/LineEnding.java
+++ b/maven-plugin/src/main/java/net/revelc/code/formatter/LineEnding.java
@@ -21,7 +21,7 @@ package net.revelc.code.formatter;
  */
 public enum LineEnding {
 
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -376,7 +375,7 @@
                         <exclude>**/src/test/resources/**</exclude>
                         <exclude>**/target/**</exclude>
                     </excludes>
-                    <lineEnding>LF</lineEnding>
+                    <lineEnding>AUTO</lineEnding>
```
velo: this ain't a good idea

AUTO for windows is different from mac that is in time different from linux.

So our build become non-deterministic

pom.xml
```diff
@@ -388,7 +387,7 @@
                 <executions>
                     <execution>
                         <goals>
-                            <goal>validate</goal>
+                            <goal>format</goal>
```
velo: I'm totally anti this change.

CI should not be formatting code.  unless CI start to push the formatted code back, which doesn't seem expected either

pom.xml
hazendaz: Nope not true.  Code is checked out as auto so it matches box.  See git attributes plus the fact so far both linux and my local windows passed.


Sent by Outlook for Android





On Wed, Oct 28, 2015 at 9:13 PM -0700, "Marvin Froeder" <notifications@github.com> wrote:
> @@ -376,7 +375,7 @@
>                          <exclude>**/src/test/resources/**</exclude>
>                          <exclude>**/target/**</exclude>
>                      </excludes>
> -                    <lineEnding>LF</lineEnding>
> +                    <lineEnding>AUTO</lineEnding>

this ain't a good idea

AUTO for windows is different from mac that is in time different from linux.

So our build become non-deterministic

---
https://github.com/revelc/formatter-maven-plugin/pull/80/files#r43347516

pom.xml
hazendaz: Ci should be validating not changing code.  That is what you said validate was for.


Sent by Outlook for Android





On Wed, Oct 28, 2015 at 9:14 PM -0700, "Marvin Froeder" <notifications@github.com> wrote:
> @@ -388,7 +387,7 @@
>                  <executions>
>                      <execution>
>                          <goals>
> -                            <goal>validate</goal>
> +                            <goal>format</goal>

I'm totally anti this change.

CI should not be formatting code.  unless CI start to push the formatted code back, which doesn't seem expected either

---
https://github.com/revelc/formatter-maven-plugin/pull/80/files#r43347551

velo: But you changed from validate to format

hazendaz:
The builds that CI enforces do not have to be enforced on a box.  Last time around we discussed you told me to use eclipse to format the code.  That isn't right.  This plugin is not required to be used in eclipse.  We use it in intellij, netbeans, eclipse, bash, etc.  Anywhere we run maven it has no tie to eclipse.  Running the build locally formats it.  On travis build, we simply have to run validate which will block it from changing line endings.

For what its worth, everything built correctly.  Including majority of even my windows build which still has issues due to the support module.  I will check that again tomorrow as it might have been resolved but it's after midnight and I'm calling it a night.

pom.xml
velo: Ow, nvm them...

But with move from LF to AUTO if all is LF since quite a while?

pom.xml
hazendaz: Absolutely.  I cannot have my local machine failing to format with the plugin.  I'm not going to use an IDE to do that.  There is a complete difference between development and CI.  If you want validate in CI, we just have to configure the travis build to run the validate.  It's really easy to do.  If it helps to see how that might work, look over at mybatis and how I ran specialized builds in travis there doing many things using bash of all things as well.  All mybatis has this now.  Here is a link to the parent.

https://github.com/mybatis/parent

There generally has to be a middle ground here.  What I need to occur on my local is automation without blocking.  What we need on CI is blocking.  I agree there.  We don't want people submitting changes unformatted which would occur otherwise.  As to any concern if CI ever was doing formatting that pretty much doesn't matter at all and never really should.  It's the source in the REPO we care about.  CI builds are rather mute.  I have this formatter running in 100+ separate projects with a major bank and that is fully accepted.  What we have an issue with is clear, code gets to master not formatted and then is formatted after the fact so that is where CI comes into play in blocking merge to master.  Even here, it's just that we force the merge to master to do the same, block unformatted code.  Pretty simple.  The key is we have to remember we don't care what peoples local repos look like.  We only care at master.  And because we can enforce travis do it, users will know way before they attempt a pull as we can block the build.

velo:
So can I do git pull, nvm clean install, git add . and git push after that?

velo:
Of course it build fine, ci will ignore any format issue, change the code and let it be

pom.xml
hazendaz: LF is unix style.  Not windows.  I realize you use a mac.  Not everyone does.  AUTO is box dependent.  Sure you can have your AUTO set to something else but that is box specific.  AUTO is on git attributes.  It has been since the original rolled into git.  That in fact is your AUTO.  It's no different.  I had nothing but issues locally with these settings.  I couldn't do anything.  I had been making changes and pushing them all up so they would build in linux.  That was the only way for me to verify a thing.  I don't have a mac and I'm not running a docker instance with linux that often.  So I relied on the externals.  I work on some 70+ projects, this is the first outside of MULE ESB that became completely unstable after merging all this code in to the point where I cannot build any longer.  If 70+ projects have no issues, then it sort of indicates a problem here.  Please take the time to look at the actual builds.  You will see that it didn't format anything.  It skipped because this is correct.  App voyer isn't doing it correctly but I fully expected that given what was going on before.  It's checking out wrong most likey as-is meaning LF and thus it will cause formatting there.  But travis is right and it's ok.

hazendaz:
travis didn't change the code.

app voyer did because it's configured to not check things out correctly to start with.  Most like as-is checkout meaning its pulling out with LF.  If I recall I thought you also mentioned app voyer wasn't working.  It is now. I didn't look back further but haven't ever used that tool either.  This setup is more in line with travis and these days that is what matters most as that is how we are really storing our source.

velo:
"app voyer did because it's configured to not check things out correctly to start with"
that is the default windows settings... like almost everyone uses... to check out the source w/o changing it

PR #81, opened 2015-10-29
Update AbstractCacheableFormatter.java

velo:
Testing build after #80


velo:
CI says all is good, looks like a good idea to merge this as it is

hazendaz:
Is this correct?  That is the source that was just applied.  The change here is to break the formatting?  GIT only shows exactly what couple of lines the other developer brought in.  It hasn't formatted since.

maven-plugin/src/main/java/net/revelc/code/formatter/AbstractCacheableFormatter.java

velo:
before PR #80 both CI servers would break, saying the format is broken.

velo:
yes, the goal of this PR is to prove PR80 introduced an error to the build.

I completely broke a file format, still the build is green.
So, either the build is wrong or we consider unformatted code a good thing

hazendaz:
that probably is the case because we have to enforce at CI level to enforce.  It's not right now.  all i did was loosen it at the development level.  at travis we will need to modify it slightly so it chokes on that.  Same with appvoyer.  that is true ci in that case.  it can be enforced in any build.  If it helps, I can work that out by end of weekend so we can then keep this PR so we can test it failing again.

hazendaz:
and then developers will still be ok to do what they want but won't be able to submit to use unless clean :)

Issue #70, opened 2015-09-07
Build fails when running on my machine


hazendaz:
Possibly due to line endings issue I mentioned on enforced validate.  The build gets this far.

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running net.revelc.code.formatter.support.io.ResourceTest
Tests run: 13, Failures: 6, Errors: 0, Skipped: 0, Time elapsed: 0.026 sec <<< FAILURE! - in net.revelc.code.formatter.s
upport.io.ResourceTest
testAsInputStreamValidDefaultPackageResource(net.revelc.code.formatter.support.io.ResourceTest)  Time elapsed: 0.019 sec
  <<< FAILURE!
org.junit.ComparisonFailure: expected:<alpha[]
]eta> but was:<alpha[
beta>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at net.revelc.code.formatter.support.io.ResourceTest.testAsInputStreamValidDefaultPackageResource(ResourceTest.j
ava:73)

testForPath(net.revelc.code.formatter.support.io.ResourceTest)  Time elapsed: 0 sec  <<< FAILURE!
org.junit.ComparisonFailure: expected:<alpha[]
]eta> but was:<alpha[
beta>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at net.revelc.code.formatter.support.io.ResourceTest.testForPath(ResourceTest.java:223)

testAsString(net.revelc.code.formatter.support.io.ResourceTest)  Time elapsed: 0.003 sec  <<< FAILURE!
org.junit.ComparisonFailure: expected:<alpha subpackage[]
]eta subpackage> but was:<alpha subpackage[
beta subpackage>
        at org.junit.Assert.assertEquals(Assert.java:115)
        at org.junit.Assert.assertEquals(Assert.java:144)
        at net.revelc.code.formatter.support.io.ResourceTest.testAsString(ResourceTest.java:114)

testAsInputStreamValidSubpackageResource(net.revelc.code.formatter.support.io.ResourceTest)  Time elapsed: 0.001 sec  <<
< FAILURE!
org.junit.ComparisonFailure: expected:<alpha subpackage[]
]eta subpackage> but was:<alpha subpackage[
beta subpackage>
        at org.junit.Assert.assertEquals(Assert.java:115)
[...]

PR #82, opened 2015-10-29
#70 Fixed - Make ResourceTest cross platform in line ending checks

hazendaz:
Fixes issue #70

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,6 +12,8 @@ branches:
   except:
     - gh-pages
 os: Windows Server 2012
+init:
+  - git config --global core.autocrlf true
 install:
   - ps: |
       Add-Type -AssemblyName System.IO.Compression.FileSystem
--- a/support/src/test/java/net/revelc/code/formatter/support/io/ResourceTest.java
+++ b/support/src/test/java/net/revelc/code/formatter/support/io/ResourceTest.java
@@ -57,6 +57,9 @@ public class ResourceTest {
 	 */
 	private static final String FILE_ROOT_TEST_FILE = "file:src/test/resources/test_file";
 
+	/** System line endings */
+	public static final String LINE_SEPARATOR = System.getProperty("line.separator");
+
 	/**
 	 * Test method for {@link Resource#asInputStream()}.
 	 * <p />
@@ -70,7 +73,7 @@ public class ResourceTest {
 		InputStream stream = forPath(ROOT_TEST_FILE).toInputStream();
 		String result = new Scanner(stream, UTF_8.name()).useDelimiter("\\A").next();
 
-		assertEquals("alpha\nbeta", result);
+		assertEquals("alpha" + ResourceTest.LINE_SEPARATOR + "beta", result);
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


velo:
this defeats the test purpose....

hazendaz:
Merging for now.  I think appveyor has issues.  It shows that it failed and contents on this break there which work on my machine.  The error though visually looks correct.  So possibly it has settings slightly different.  Each test that failed appears to be \r\n which is exactly as expected so unsure why it chocked on that but given it was failing prior to other changes and now is having issues I think some more looking there is needed.  I still heavily rely on travis for being correct though as that more closely represents the source given we store it the linux way.

velo:
Breaks for me too!

I think your env has issues

hazendaz:
I'll look at appveyor tomorrow.  I would suggest you also check your git settings.  I've seen issues like this all of twice in 4 years from two people.  One being yourself and another being one from a team in indea.  It's almost always an incorrect setting in how line endings are dealt with.  When installing GIT (I don't recall the specific config setting), make sure the correct setting for line endings is selected.  This library is setup for GITHUB and uses their settings by default as does every other project I've worked on.  I'll hang on this PR until after I look more at appveyor.  If it's anything like cloudbees, its probably got issues or incorrect setup.  If travis works and my machine works, it's good.  I work far too many repos and commit far too much for this to be an issue only on this project.  I don't have line ending issues on any other projects.  Using '\n' in a test is really bad as that is platform setting specific.  It's always best to use System.getProperty("line.endings") in that case.  That ensures it picks it from the system.

hazendaz:
Prior builds were failing rather than this one.  There is some linkage issue of sorts from app veyor where it tagged master build against this PR.  .gitattributes did successfully handle this correctly as expected by checking out correctly in windows.  On top of that, I've now added setting to appveyor.yml file which doubly ensures we have this forced correctly.  I watched this PR build without issue both ways.  I further found documentation with node.js stating specifically how appveyor does things and to never rely on their settings but rather be explicit and/or setup in gitattributes as we already had set.  This was good days ago.  In fact, this is what fixed the master.  Now it's got twice the protection.  Additionally, having this fixed means that it does not matter what anyone else does in windows at all.  It's no longer system dependent on a line ending style which was linux only.  Now it will work regardless of setup.  Case closed!

appveyor.yml
```diff
@@ -12,6 +12,8 @@ branches:
   except:
     - gh-pages
 os: Windows Server 2012
+init:
+  - git config --global core.autocrlf true
```
velo: so also include this special instructions on README.md, cause otherwise, people that try to build this project w/o magic configuration will fail.

which still is a bad idea

hazendaz:
there isn't any magic configuration here.  If you don't think so, then I highly suggest doing some reading on line endings and getting a cross platform understanding and further learning why you should never write unit tests that check specifically for '\n'.  Even in source code, that is a huge no-no in java as it will practicly ensure code doesn't run on multiple environments.

velo:
oddly enough, test worked across platform, w/o configurations before you 'fixed' it.

PR #119, opened 2016-08-06
Support multimodule projects

elrodro83:
With this change, a formatter config file may be present in just the root project in a maven multimodule, and having the submodules reference that file.

Without this change, when configuring the formatter for a submodule, it failed if the referenced formatter config was outside the submodule (i.e.: in the root project).

Also, added the fileName when it fails for troubleshooting.

### DIFF
--- a/m2e-configurator/net.revelc.code.formatter/src/net/revelc/code/formatter/connector/internal/FormatterProjectConfigurator.java
+++ b/m2e-configurator/net.revelc.code.formatter/src/net/revelc/code/formatter/connector/internal/FormatterProjectConfigurator.java
@@ -17,6 +17,8 @@
 package net.revelc.code.formatter.connector.internal;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


m2e-configurator/net.revelc.code.formatter/src/net/revelc/code/formatter/connector/internal/FormatterProjectConfigurator.java
```diff
@@ -125,14 +126,20 @@ private InputStream readConfigFile(Formatter formatter, ProjectConfigurationRequ
             javaConfigFile = formatter.getDefaultPath();
         }
 
-        IFile cfgFile = request.getProject().getFile(javaConfigFile);
-
-        if (!cfgFile.exists()) {
-            throw new CoreException(
-                    new Status(IStatus.CANCEL, FormatterCore.PLUGIN_ID, "Configuration file not found!"));
+        File cfgFile = new File(javaConfigFile);
+        if (!cfgFile.isAbsolute()) {
```
velo: so, if the file exists we will look for another one that may or may not exists?!

:-1:

hazendaz: I think all this is doing different is checking to see if an absolute path exists.  If it does, it does the same as it always did.  If it does not, then it performs the additional check.  So worst case this is just adding something, best case it does as stated.  As I don't use the m2 configuration not sure how this completely works.  I did try to go to the core plugin and implement something similar but that didn't really do anything.  I did however get full multi module support so format file can be in parent and no matter how deep the maven build is, it will recursively search.  Also of note, plexus utils got much smarter so old lost logic of jar not working now works.  So we should be relatively close to release at this point.  As for this change, is this something you can test based on what user documented?

velo: this doesn't check about existence...

```new File("file.txt").isAbsolute(); // false```
```new File("file.txt").exists(); // may be true or false```

```new File("/file.txt").isAbsolute(); // true```
```new File("file.txt").exists(); // may be true or false```

so being absolute has nothing to do with existence

so the file may exist and not be absolute...  what I just don't know is if this ```request.getProject().getLocation().append()``` will be able to find the existing relative path file

Issue #123, opened 2016-08-08
Formatter attempts to format readonly files


fbyrne:
The formatter does not skip files readonly files that cannot be formatted.

This is important for SCMs which require checkout for example, perforce.

hazendaz:
Closing as fixed in #129

PR #124, opened 2016-08-08
Fixes Issue #123 Formatter attempts to format readonly files (#1)

fbyrne:
* Block formatting of non writable files.  

This is for scm's that require files to be checked out before editing.  For example, perforce.

* Update FormatterMojo.java

### DIFF
--- a/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -280,7 +280,7 @@ public class FormatterMojo extends AbstractMojo implements ConfigurationSource {
             String basedirPath = getBasedirPath();
             for (int i = 0, n = files.size(); i < n; i++) {
                 File file = files.get(i);
-                if (file.exists()) {
+                if (file.exists() && file.canWrite()) {
                     formatFile(file, rc, hashCache, basedirPath);
                 }
             }
@@ -319,7 +319,10 @@ public class FormatterMojo extends AbstractMojo implements ConfigurationSource {
 
         List<File> foundFiles = new ArrayList<>();
         for (String filename : ds.getIncludedFiles()) {
-            foundFiles.add(new File(newBasedir, filename));
+            File foundFile = new File(newBasedir, filename);
+            if(foundFile.canWrite()){
+                foundFiles.add(foundFile);
+            }
         }
         return foundFiles;
     }


maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
```diff
@@ -280,7 +280,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             String basedirPath = getBasedirPath();
             for (int i = 0, n = files.size(); i < n; i++) {
                 File file = files.get(i);
-                if (file.exists()) {
+                if (file.exists() && file.canWrite()) {
```
hazendaz: @fbyrne I agree this fixes the issue....

maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
```diff
@@ -319,7 +319,10 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
         List<File> foundFiles = new ArrayList<>();
         for (String filename : ds.getIncludedFiles()) {
-            foundFiles.add(new File(newBasedir, filename));
+            File foundFile = new File(newBasedir, filename);
+            if(foundFile.canWrite()){
+                foundFiles.add(foundFile);
+            }
```
hazendaz: @fbyrne However, this fixes it before it ever becomes an issue.  Given the cost of creating objects in loop and fact that we are already doing this in the other location, I think it would be cheaper to just of the first fix.  Thoughts?

maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
hazendaz: @fbyrne I'm going to pull this in a separate pull request to master shortly.  I've enhanced it further so we get a count of files that were tagged read only.  This is certainly better than throwing tons of stack traces and will help developers as well as some scm(s) such as mks integrety require same checkout as you noted.

hazendaz:
Closing for now.  First portion of fix that required little to no change to code base accepted via #129.  The second portion seems like duplication and causes extra object creation that probably isn't worth having.  If submitter has strong desire for the original change and can persuade us to change, we can re-address in a separate pull request.  Thanks for your contribution @fbyrne

maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
fbyrne: I didn't have a huge amount of time to do the fix.  I had originally added the fix before the call to formatFile(), however, this made the file counts inconsistent.  So I added the check in addCollectionFiles() to prevent readonly files from being added.  For completeness, I left the check around formatFile as if the file list is large there is a chance the status of the file can change from creation of the list to formatting of the file.

fbyrne:
No strong desire, Ill likely propose some further changes for performance on large codelines in the coming week.

hazendaz:
There was a bug in the counts on general.  I fixed that and also added new count for read only files.

On Tue, Aug 9, 2016 at 4:51 AM -0400, "Fergus Byrne" <notifications@github.com<mailto:notifications@github.com>> wrote:


No strong desire, Ill likely propose some further changes for performance on large codelines in the coming week.

-

PR #130, opened 2016-08-09
[config] Add support for multimodule build with file in parent only

hazendaz:
This code supports multimodule builds by allow a single format xml file in the root directory.  It will recursively read back until there are no more root pom(s) before it stops looking for the first format xml it comes across.  Tested this via [Waffle](https://github.com/dblock/waffle) as it has many layers and was well suited to test this type of usage as it is a bit complex.  Logging exists in debut to show what is going on during searching.

### DIFF
--- a/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -28,6 +28,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -562,6 +563,17 @@ public class FormatterMojo extends AbstractMojo implements ConfigurationSource {
 
         this.getLog().debug("Using search path at: " + this.basedir.getAbsolutePath());
         this.resourceManager.addSearchPath(FileResourceLoader.ID, this.basedir.getAbsolutePath());
+
+        // Add parents
+        File parent = new File(Paths.get(this.basedir.toURI()).getParent() + File.separator + "pom.xml");
+        this.getLog().debug("Probing pom at: " + parent.getAbsolutePath());
+        while (parent.exists()) {
+            this.getLog().debug("Using search path at: " + parent.getParent());
+            this.resourceManager.addSearchPath(FileResourceLoader.ID, parent.getParent());
+
+            parent = new File(Paths.get(parent.toURI()).getParent().getParent() + File.separator + "pom.xml");
+            this.getLog().debug("Probing pom at: " + parent.getAbsolutePath());
+        }
 
         try (InputStream configInput = this.resourceManager.getResourceAsInputStream(newConfigFile)) {
             return new ConfigReader().read(configInput);
--- a/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -213,6 +213,15 @@ public class FormatterMojo extends AbstractMojo implements ConfigurationSource {
     @Parameter(defaultValue = "false", alias = "skip", property = "formatter.skip")
     private Boolean skipFormatting;
 
+    /**
+     * Add configuration search for all parents of module executing formatting. The
+     * first found configuration including in module will be used.
+     * 
+     * @since 1.6.0
+     */
+    @Parameter(defaultValue = "false", property = "addConfigSearchInParents")
+    private Boolean addConfigSearchInParents;
+
     private JavaFormatter javaFormatter = new JavaFormatter();
 
     private JavascriptFormatter jsFormatter = new JavascriptFormatter();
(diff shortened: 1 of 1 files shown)


ctubbsii:
I'm not sure about this recursion. This isn't a normal maven way of doing things. Usually, for multi-module projects, you'd provide the formatter config file in a build-resources jar, which is added as a dependency to the plugin, so it's on the plugin's classpath. Or, you'd set the config file in the parent to be relative to the children as in `<config>../src/main/build-resources/formatter.xml</config>` and have the plugin fail gracefully when it can't find the file when the parent itself executes.

Granted, these conventions aren't necessarily ideal for multi-module projects, but is this recursion behavior sufficiently better to warrant breaking conventions? Or, is there precedent for this kind of recursion?

hazendaz:
There is a plugin that can find the root location which essentially probably does a bit of the same internally.  I didn't look there but the plugin is so old and requires special knowledge of how to set it up which is enforced on user.  Otherwise, exactly as you state.  Personally, I use the Jar solution. 

However, have seen others wanting to do something like this with less complex setup (ie less configuration).  This sort of fits the bill and as you can imagine took a bit to get it figured out.  This works with both execution in module as well as parent which is generally a headache when manually configuring such things.  Essentially this ends up adding all potential places to look in order to look (module/parent/more parents).  That is done based on pom.xml existing not the actual formatter.  Then maven does the rest to look in those locations one after another.  If those fail, it uses it's other ways to get there.

Now if maven ever addresses providing a variable that shows the true root location, we would by far have easiler solution but then again this is a lot more flexible in that you actually could run different formatters at various levels if you so choose to do so.

ctubbsii:
Maven does provide `${project.parent.basedir}` and `${project.parent.build.directory}`, and you can also do `parent.parent` to get the grandparent, so you don't have to assume it's one directory up `../` (though I'd expect any multi-module project to be aware of its own particular layout).

In any case, the argument that this is less complex (less configuration) is convincing.

I've used both the jar strategy and the `${project.parent.basedir}` strategy. Since the parent POM in a multi-module build in all my cases is `<packaging>pom</packaging>`. The latter is pretty simple configuration, since the formatter doesn't run on that anyway because the plugin's execution is not bound to any phase which exists in the `pom` packaging lifecycle by default (it's bound to `process-sources` by default, which doesn't exist in the pom lifecycle).

ctubbsii:
If anybody is interested, a good example of a multi-module use of this plugin is here: https://github.com/apache/accumulo/blob/master/pom.xml

hazendaz:
That one uses the jar if I understand it correctly.  Same thing I do at my day job.  In my module, I use jar at a super pom level, then to not impose formatting right away on users, have them drop in an empty format.xml file when they are ready.

ctubbsii:
No. Accumulo ships the config file in the contrib directory at the root. Here's an example with the jar: https://github.com/apache/incubator-fluo/blob/fluo-parent/pom.xml

hazendaz:
Thanks ill check both out.

On Tue, Aug 9, 2016 at 12:25 AM -0400, "Christopher Tubbs" <notifications@github.com<mailto:notifications@github.com>> wrote:


No. Accumulo ships the config file in the contrib directory at the root. Here's an example with the jar: https://github.com/apache/incubator-fluo/blob/fluo-parent/pom.xml

-

hazendaz:
So looking at the project with configuration in parent,  parent Parent only gets it back to the poms parent and since that uses relative directory back two it works.  The test project I used has many levels with multiple parents so this would indeed solve other needs.  However, as a compromise, how about I add this with a flag so it's by default not even on but if those that might want it, they can turn the flag on?

Personally I feel maven should support something like this right out of the box as I've seen this many times asked for and still to this day no supported maven solution but that is a bigger issue in general.  There is one plugin that does address it but it's really old and probably not on maven teams radar.  At this time though I don't have the time to chase down a new feature in the maven core libraries which may never happen.  And I only personally have about maybe 3 use cases where this would be nice-to-have feature.

ctubbsii:
The way accumulo used to support deeper levels was to override:

```xml
<eclipseFormatterStyle>${project.parent.basedir}/contrib/Eclipse-Accumulo-Codestyle.xml</eclipseFormatterStyle>
```

with:

```xml
<eclipseFormatterStyle>${project.parent.parent.basedir}/contrib/Eclipse-Accumulo-Codestyle.xml</eclipseFormatterStyle>
```

And the config for the formatter just referenced this property. When the plugin is inherited, the config is evaluated in the child, and this property is substituted.

However, that still requires the multi-module project owner to make adjustments to the pom, based on the project's hierarchy depth. I agree this would be better if Maven just supported this out of the box. Something like: `${project.root.basedir}`

I'm not strongly against doing this in this plugin, but I personally prefer another option, suitable for all plugins. I just created https://github.com/mojohaus/build-helper-maven-plugin/issues/29 because I think it's a better alternative than proceeding with this feature, or waiting for Maven to add it themselves.

If you do proceed, I don't personally have an opinion on whether it's on by default or not, so long as it's configurable, so we could turn it off if necessary.

hazendaz:
@ctubbsii I updated this with configuration default to false.  I've tested it's working.  Can you review and possibly give suggestions as to propery naming?  I'll squash this before bringing it in so it's a single commit.

ctubbsii:
FWIW, build-helper-maven-plugin will soon have the ability to set a property to refer to the root, which will make it easier for this, and any other plugin to find root-level resources from child modules: https://github.com/mojohaus/build-helper-maven-plugin/commit/93acd4ed714e89a6eb5f0ad7ef5d058d839e2b29

Once that is released, I don't think we'll need this change to the formatter.

***

I looked at the current code change. I haven't been able to test it, but I think the property name should have the "formatter." prefix, and might be bettter named: `formatter.searchParentsForConfig`

Is this recursive behavior the desired behavior, or should we just search the execution root? If the latter, the build-helper code linked above might be a good demonstration for directly adding the execution root.

hazendaz:
Closing for now.  The recursive nature isn't necessary in this just happened to be how it worked out.  I'll try the build helper one once it's available and if it solves the issue I think that will be good enough.  I'll save this off for now in case we want to explore it more later on.

hazendaz:
@ctubbsii The logic added to build-helper-maven-plugin doesn't quite work properly.  It works only if running from the root of a project.  It fails otherwise.  I have it on a couple of projects and wasn't too pleased with that outcome :(

ctubbsii:
@hazendaz That's not too surprising. Maven multi-module projects are designed to run from project root, and many plugins break if you don't follow that convention. The feature in buildhelper was designed with that in mind.

PR #156, opened 2016-12-29
[pom] Switch to using self -1 for internal formatting requirements

hazendaz:


hazendaz:
Bla...we might have a bug in the code that @velo handled better.  I think there was an open ticket on it so probably what he was mentioning long ago but the flip side broke code I had.  So something to look at.  At any rate, switched how it gets the main basedir to run during validate phase and run that during travis before script.  This also highlighted since I moved this to wrapper I needed those to use the wrapper for maven.  Configuration was moved to dependency management for this piece as well.  It's working now as expected other than need to define the two locations for the config.

PR #183, opened 2017-07-01
Mark mojos as thread-safe

elrodro83:
Since the mojos don't change the default instantiation strategy, nor they have static non-final attributes, they are thread safe.
They should be marked so to prevent maven from logging warning when using the plugin on parallel reactor builds.


hazendaz:
@elrodro83 Thanks

PR #201, opened 2017-09-24
[travis] Add openjdk8

hazendaz:


ctubbsii (approved):
Surprised that works. Last time I tried, this want an option.

hazendaz:
It just changed recently.  Trusty builds of travis are default now and drop openjdk6, oraclejdk7, and added openjdk8.

PR #205, opened 2017-09-29
Fix site deploy fixes #188

### LINKED ISSUE #188: Site is not intact
While the main site page is showing todays date, the remainder is still showing from 2015.  Thus far I did a full maven build with site and then a deploy.  All that did was fix the first landing page.  The underlying pieces like usage are showing the old dates.  We need to figure this out so users have proper documentation.

hazendaz:
Fixes #188


hazendaz:
Site is already deployed and from root all is well with current information rather than dropping back to our old builds from 2 years ago.

PR #209, opened 2017-10-03
Takari 5 lifecycles as one was causing issues

hazendaz:
We were not actually building correctly at all since July.  This was unusable code best I can determine.  We don't use takari ourselves but tycho is and it's enforcing upon us.  Turned it off and added additional comments which align with 5 plugins it wants to replace.  The code is now usable again.


hazendaz:
Removing test fixes.  Seem system dependent.  Will review that again later.  Just fixing takari now.

PR #215, opened 2017-10-06
[java6] Raise compiler default levels to 1.6

hazendaz:
Java 5 is really dated now.  We should start with java 6 compliancy.  User can override if they need to.


hazendaz:
Adding formatting cleanup here to avoid unnecessary additional PR.

PR #217, opened 2017-10-06
[improve] Bring back default formatting with compiler level settings …

### LINKED ISSUE #216: Default Formatting Java/Javascript
Original code had ability to run default formatting with eclipse with no config file and setting the compiler level for the formatters to use by default.  This was lost with @velo indicating problem in original code base merges to pickup javascript and eclipse libaries from him.  At that time we had no real solution due to some diverging usage.  Through research, it was determined it currently is not possible to hit the following code in mojo formatter as there is always a value in the file to lookup (may not even exist).  This would have been the default code had it been working.

in getFormattingOptions
```
        Map<String, String> options = new HashMap<>();
        options.put(JavaCore.COMPILER_SOURCE, this.compilerSource);
        options.put(JavaCore.COMPILER_COMPLIANCE, this.compilerCompliance);
        options.put(JavaCore.COMPILER_CODEGEN_TARGET_PLATFORM, this.compilerTargetPlatform);

        return options;
```

Because we were populating the map immediately back this was entirely skipped.  Worse still, we where throwing an exception if there was no file to process configuration from.  There was a failsafe catch though on java only.  However, that would have been bypassed as well because the options were flagged null after #109 and the formatting in that case entirely skipped.

So to fix this, there was no need to have the failsafe catch on the java side.  Also, it made no sense to check that file attempting to be read was null as it is always present.  The fix will simpy try to get the options map.  If that is not found it will now send empty back.  At that point, it will then add the options (compiler source, compiler compliance, and complier codegen target platform).

While this really cleans up the code, it needs a review from others.  More importantly, need @velo to assist on this to make sure this makes sense for both java and javascript.  The options put here is out of javaCore but based on research it technically should have set in this area even for javascript although that might not actually make sense.  I will leave this issue and the fix soak for a few days to get feedback.

hazendaz:
…through use of new parameter

fixes #216

### DIFF
--- a/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/maven-plugin/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -649,14 +649,14 @@ public class FormatterMojo extends AbstractMojo implements ConfigurationSource {
      * @throws MojoExecutionException the mojo execution exception
      */
     private Map<String, String> getFormattingOptions(String newConfigFile) throws MojoExecutionException {
-        if (newConfigFile != null) {
-            return getOptionsFromConfigFile(newConfigFile);
-        }
+        Map<String, String> options = getOptionsFromConfigFile(newConfigFile);
 
-        Map<String, String> options = new HashMap<>();
-        options.put(JavaCore.COMPILER_SOURCE, this.compilerSource);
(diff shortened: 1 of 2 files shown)


hazendaz:
8 days, I'm comfortable with this.  No reviews so merging.  Merging.

PR #232, opened 2017-10-22
[ci] Move jsoup into it's own package

hazendaz:


hazendaz:
Ignoring issue on travis at the moment as this is unrelated.  I am looking at that but it's inconsistent failure at the moment.

PR #246, opened 2017-10-26
Use eclipses jdt-core directly and disable progard for jsdt core as not fully implemented

hazendaz:
Use the official jdt-core from eclipse.  For eclipse Neon which we are targetting that is 3.12.3.

Progard intentions were good but it wasn't actually used by the plugin at all nor was it complete.  Various attemps to adjust it ran into issues with illegal access presumably because it might have been mixing up classes to some degree.  If thee is a strong need for this, it is gone with moving to the official jdt-core because it was for size savings and we clearly would be giving that up.  However, if really needed, why not use felix plugin instead since it's more osgi like to start with.

This PR is more for review than anything but can be merged.  A follow-up would officially remove all the pieces we are intended to get rid of.  The ending structure would stay the same in case we need to pull anything else in from eclipse or elsewhere.

### DIFF
--- a/formatters/pom.xml
+++ b/formatters/pom.xml
@@ -30,7 +30,8 @@
     <packaging>pom</packaging>
 
     <modules>
-        <module>jdt-core</module>
+        <!-- No longer necessary -->
+        <!-- <module>jdt-core</module> -->
         <module>jsdt-core</module>
     </modules>
 
--- a/formatters/pom.xml
+++ b/formatters/pom.xml
@@ -61,6 +61,8 @@
                             </execution>
                         </executions>
                     </plugin>
+                    <!-- Unnecessary inclusion never actually used -->
+                    <!-- 
                     <plugin>
                         <groupId>com.github.wvengen</groupId>
                         <artifactId>proguard-maven-plugin</artifactId>
@@ -85,6 +87,7 @@
                             </execution>
                         </executions>
                     </plugin>
+                    -->
                 </plugins>
             </build>
         </profile>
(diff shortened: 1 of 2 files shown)


ctubbsii (approved):
I think this change is good. But, I think it's safe to just rip this stuff out, rather than leave it commented.

Also, why use version 3.12.3 instead of the latest in Maven Central, 3.13.0?

hazendaz:
Only reason to use that version is if we continue to target eclipse versions to ensure consistent formatting.

ctubbsii:
It's kind of hard to tell which versions match up to a particular major version of Eclipse. I don't know that we need to continue to target the major (named) versions, but I had just assumed that 3.13.0 was just a newer version of Neon, but now that I think about it, it's probably Oxygen. We can target Oxygen in a 3.0 release.

I'm currently working on separating out jsdt-core as a separate revelc project, so it should help us do releases a bit better without that. I'm still not sure about the m2e-configurator modules... as I don't yet fully understand how those work, but at least we won't be shipping EPL-licensed stuff in the AL2-licensed formatter-maven-plugin builds any more.

I also found a way to bump the OSGi manifest version during execution of the maven-release-plugin, when it increments the version in the POM. So, hopefully that will also help with releases, if we still need the Tycho builds for the m2e-configurator.

hazendaz:
Sounds Great!

Probably the easiest way to see what version of Eclipse is to check eclipse installation details.  For .3 version of neon, it was 'Eclipse Java Development Tools 3.12.3.v20170301-0400.  And yes 3.13.0 is oxygen.  They seem to be quite slow though to getting those into central.  Oxygen is already on 3.13.1 but 3.13.0 just got to central a couple of weeks ago.  I'm hoping they keep those coming.

Issue #266, opened 2018-01-16
"Class or interface declaration" (org.eclipse.jdt.core.formatter.brace_position_for_type_declaration) does not work with "next line"


smatei:
I have set the "Class or interface declaration" formatting preference to "next line" in eclipse, and exported it to a multi-module "build-tools" src\main\resources\eclipse\formatter.xml. When compiling the project with maven I get the following output, that tells me that my file Test.java is formatted (I am on Windows 10 64 bits, and use JDK build 1.8.0_151-b12 on 64 bits, with maven 3.5.0).

> 		
		$ mvn clean compile
		[INFO] Scanning for projects...
		[INFO] ------------------------------------------------------------------------
		[INFO] Reactor Build Order:
		[INFO]
		[INFO] all
		[INFO] utils
		[INFO]
		[INFO] ------------------------------------------------------------------------
		[INFO] Building all 0.0.1-SNAPSHOT
		[INFO] ------------------------------------------------------------------------
		[INFO]
		[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ all ---
		[INFO]
		[INFO] --- formatter-maven-plugin:2.7.1:format (default) @ all ---
		[INFO] Using 'UTF-8' encoding to format source files.
		[INFO] Number of files to be formatted: 0
		[INFO]
		[INFO] ------------------------------------------------------------------------
		[INFO] Building utils 0.0.1-SNAPSHOT
		[INFO] ------------------------------------------------------------------------
		[INFO]
		[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ utils ---
		[INFO] Deleting D:\_git\parent\utils\target
		[INFO]
		[INFO] --- formatter-maven-plugin:2.7.1:format (default) @ utils ---
		[INFO] Using 'UTF-8' encoding to format source files.
		[INFO] Number of files to be formatted: 1
		[INFO] Successfully formatted:          1 file(s)
		[INFO] Fail to format:                  0 file(s)
		[INFO] Skipped:                         0 file(s)
		[INFO] Read only skipped:               0 file(s)
		[INFO] Approximate time taken:          0s
		[INFO]
		[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ utils ---
		[WARNING] Using platform encoding (Cp1252 actually) to copy filtered resources, i.e. build is platform dependent!
		[INFO] skip non existing resourceDirectory D:\_git\parent\utils\src\main\resources
		[INFO]
[...]

PR #267, opened 2018-01-19
Multimodule configuration example - fixed confusing dependency.

smatei:
Example was confusing, point to the build-tools module defined by the user in the parent pom.

[Issue 266](https://github.com/revelc/formatter-maven-plugin/issues/266)

If I define a build-tools module in my multiproject-parent, I want it as a dependency for the formatter-maven-plugin.

### DIFF
--- a/src/site/apt/examples.apt.vm
+++ b/src/site/apt/examples.apt.vm
@@ -282,9 +282,9 @@ multiproject
         </configuration>
         <dependencies>
           <dependency>
-            <groupId>com.github.hazendaz</groupId>
+            <groupId>com.googlecode.example.multiproject</groupId>
             <artifactId>build-tools</artifactId>
-            <version>1.1.2</version>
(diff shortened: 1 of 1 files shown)


hazendaz:
How does changing example to show another clarify anything?  I think describing it might help others more.

For note, build tools is mine in the example.  It's a single module project released to central used by my own base parent also in central and both repos are on github (github.com/hazendaz).  It's also used in mybatis and a few others.  Hope that helps.

ctubbsii:
Changing to a package which doesn't exist might be better than the current package which does exist, but which does not contain the specified formatter config. The failure to resolve the non-existent artifact will make it more obvious that this is just an example.

However, I think we can do a little better.

ctubbsii (approved):
This makes the example slightly more consistent with the example documentation, but I think we can do better. I'll merge this in and then try to improve the wording a bit.

PR #277, opened 2018-05-26
create a multi-module project to support ant and maven both.

divyakumarjain:
The PR address #278. This is a WIP code.


divyakumarjain:
I like the simplicity of the project and the structure of the project. This simplicity allowed quick conversion to multi-module  project. For which the Java code was not changed. The only thing which was changed was the logging library.

The Maven structure was changed to have common code in core module which can be shared by both Maven and Ant.

Thus the simplicity of the java code did not changed a lot but yes maven structure was altered to make sure common pieces are shared between ant and maven.

I think it boils down to simplicity of maven structure vs duplication of code. Let me know if a different approach is preferred where common code can be shared between ant and maven so that any new feature can be added to both quickly without duplication of the code. I will not be able to contribute long term to this project so you guys will be best judge for this.

ctubbsii:
If the only incompatibility is the logger, then we could probably just switch to using the SLF4J logger, since Maven is moving to that anyway (https://maven.apache.org/maven-logging.html) over the Plexus logger. But, I'm still confused as to why any Ant project would want to use a Maven plugin wrapper to execute these tools, when they could just use these tools directly.

ctubbsii:
Perhaps I could understand better if you had an example Ant project you could link to which provided a simple demonstration for how it works now in Ant, and how you'd like it to work?

divyakumarjain:
Yes, we can move to SLF4J.

Ant will not use Maven plugin but it will only use core classes. Please see below for explanation.

image

ctubbsii:
This project is specifically a "Maven" plugin, and I really don't think it makes sense to turn it back into a multi-module project to support non-Maven use cases... since its sole purpose for existing is to run during a Maven build.

The code in these "core" classes aren't really doing the real formatting work... they're just Maven wrappers around the real workers, which are the project's dependencies (like `jdt-core` and `jsdt-core`), and those are already separate projects you can presumably use in Ant.

However, I do think there are things we can do to help you use it in your Ant project, if that would be helpful, while keeping the project in a single-module. We can move the Mojos into their own package, and create classes which can be instantiated directly, which do not rely on any Maven framework classes. This is basically the strategy taken with the `impsort-maven-plugin`.

We can also probably fix the logging, so these classes don't use the Mojo's logger.

Please allow me some time to make an attempt at such a refactor, to see if it will suffice for your use case. I will create a separate pull request, and link to this one with a mention.

divyakumarjain:
Let's do it as you suggested. Also personally I will not be able to invest much time on this issue for a week.

Let me know once the refactoring is done i will be happy to be a beta tester for this.

divyakumarjain:
@ctubbsii Any updates on this topic. I am still looking for a Ant way to format Java source code.

ctubbsii:
@divyakumarjain Sorry, it's been near the top of my TODO list for some time, but I haven't managed to get to it yet. I had some personal/family stuff come up recently. I'll see if I can revisit this soon.

hazendaz:
@divyakumarjain @ctubbsii It's been another year at this point.  Ant is barely a competiter and isn't that well maintained any longer as in the past.  Plus maven can run ant so even if you needed to use this project, you could simply have 'ant' call maven tasks and run this.  I opt that we close this and the associated issue instead.

Some stats to help see market analysis.  This is just one, but pretty much everywhere 'ant' is long out.  I think we would have more grounds for supporting other options like gradle but even then that can still turn around and run maven so while not as straight forward a setup, I don't see any of us making any movement here.

https://www.datanyze.com/market-share/build-auto--317/apache-ant-vs-apache-maven

ctubbsii:
I concur with @hazendaz . Let's close this for now. If a demand resurges with a stronger case for a refactor, this can be revisited.

PR #298, opened 2019-03-26
Adds xml-formatter

jam01:
Replaces jsoup with revelc's [xml-formatter](https://github.com/revelc/xml-formatter)

### DIFF
--- a/src/main/java/net/revelc/code/formatter/xml/XMLFormatter.java
+++ b/src/main/java/net/revelc/code/formatter/xml/XMLFormatter.java
@@ -19,6 +19,8 @@ import net.revelc.code.formatter.AbstractCacheableFormatter;
 import net.revelc.code.formatter.ConfigurationSource;
 import net.revelc.code.formatter.Formatter;
 import net.revelc.code.formatter.LineEnding;
+import net.revelc.code.formatter.xml.lib.FormattingPreferences;
+import net.revelc.code.formatter.xml.lib.XmlDocumentFormatter;
 
 /**
  * @author yoshiman
--- a/src/main/java/net/revelc/code/formatter/xml/XMLFormatter.java
+++ b/src/main/java/net/revelc/code/formatter/xml/XMLFormatter.java
@@ -39,9 +39,9 @@ public class XMLFormatter extends AbstractCacheableFormatter implements Formatte
         String splitMultiAttrs = options.get("splitMultiAttrs");
 
         prefs.setMaxLineLength(maxLineLength != null ? Integer.valueOf(maxLineLength) : null);
-        prefs.setTabWidth(tabWidth != null ? Integer.valueOf(tabWidth) : null);
         prefs.setWrapLongLines(wrapLongLines != null ? Boolean.valueOf(wrapLongLines) : null);
         prefs.setTabInsteadOfSpaces(tabInsteadOfSpaces != null ? Boolean.valueOf(tabInsteadOfSpaces) : null);
+        prefs.setTabWidth(tabWidth != null ? Integer.valueOf(tabWidth) : null);
         prefs.setSplitMultiAttrs(splitMultiAttrs != null ? Boolean.valueOf(splitMultiAttrs) : null);
 
         this.formatter = new XmlDocumentFormatter(options.getOrDefault("lineending", System.lineSeparator()), prefs);
--- a/src/main/java/net/revelc/code/formatter/xml/XMLFormatter.java
+++ b/src/main/java/net/revelc/code/formatter/xml/XMLFormatter.java
@@ -13,19 +13,54 @@
  */
 package net.revelc.code.formatter.xml;
 
+import java.util.Map;
+
+import net.revelc.code.formatter.AbstractCacheableFormatter;
+import net.revelc.code.formatter.ConfigurationSource;
 import net.revelc.code.formatter.Formatter;
 import net.revelc.code.formatter.LineEnding;
-import net.revelc.code.formatter.jsoup.JsoupBasedFormatter;
 
 /**
  * @author yoshiman
- *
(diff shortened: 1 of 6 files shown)


jam01:
hmm I the xml formatter takes in line ending as part of configuration options, but not when invoking `format(String)` I guess I'll modify the test to assert different results like it was before.

src/main/java/net/revelc/code/formatter/xml/XMLFormatter.java
```diff
@@ -13,19 +13,54 @@
...
+        String tabWidth = options.get("tabWidth");
+        String splitMultiAttrs = options.get("splitMultiAttrs");
+
+        prefs.setMaxLineLength(maxLineLength != null ? Integer.valueOf(maxLineLength) : null);
+        prefs.setTabWidth(tabWidth != null ? Integer.valueOf(tabWidth) : null);
+        prefs.setWrapLongLines(wrapLongLines != null ? Boolean.valueOf(wrapLongLines) : null);
+        prefs.setTabInsteadOfSpaces(tabInsteadOfSpaces != null ? Boolean.valueOf(tabInsteadOfSpaces) : null);
+        prefs.setSplitMultiAttrs(splitMultiAttrs != null ? Boolean.valueOf(splitMultiAttrs) : null);
+
+        this.formatter = new XmlDocumentFormatter(options.getOrDefault("lineending", System.lineSeparator()), prefs);
```
hazendaz: @jam01 Is 'lineending' case sensitive here and possible needs to be 'lineEnding'?

Is it also possible to put order of strings to match order of the preference loading?  That makes review a bit easier :)

jam01: It's 'lineending' because it's supposed to be like json's formatter which configures it when instantiating the formatter, but your comment uncovered that it wasn't setup correctly from the mojo, so thanks.

Solved the order of the strings :+1:

hazendaz: Perfect!  Thanks.

src/test/java/net/revelc/code/formatter/xml/XMLFormatterTest.java
```diff
@@ -24,18 +24,21 @@
...
 /**
  * @author yoshiman
+ * @author jam01
  */
 public class XMLFormatterTest extends AbstractFormatterTest {
 
     @Test
     public void testDoFormatFile() throws Exception {
-        // FIXME Handle linux vs windows since this formatter does not accept line endings
+        // Since we set the line endings via options for xml, we cannot rely on CRLF inside doTestFormat.
```
hazendaz: @jam01 Trivial thing here.  Can you note xml formatter rather than json formatter (copy/paste gotcha)?

jam01: :heavy_check_mark:

jam01:
Good to merge as-is?

src/main/java/net/revelc/code/formatter/xml/XMLFormatter.java
```diff
@@ -13,19 +13,54 @@
...
 import net.revelc.code.formatter.LineEnding;
-import net.revelc.code.formatter.jsoup.JsoupBasedFormatter;
 
 /**
  * @author yoshiman
- *
+ * @author jam01
  */
-public class XMLFormatter extends JsoupBasedFormatter implements Formatter {
+public class XMLFormatter extends AbstractCacheableFormatter implements Formatter {
```
ctubbsii: One thing that could be an issue that I hadn't realized before: the package name here conflicts with the xml-formatter library. In general, this is a bad idea, and could prevent the JVM from loading classes if they are packaged in sealed jars.

jam01: Hmm was not aware of that. Do you have a change in mind? I suppose the package names should change on the xml-formatter project, since this follows the convention on this project.

ctubbsii: Yeah, we can change it there.

ctubbsii: Released the package-renamed version as 0.1.1; should sync to Maven Central within a few hours.

jam01:
Thanks again @ctubbsii . I'm a mule developer and mule application code is almost exclusively XML, similar to Apache Camel. A way to automatically keep the XML code formatted should be really helpful. I'm not familiar with xmllint but this formatter plugin seemed like a great option.

ctubbsii:
For what it's worth, `xmllint` is part of libxml/libxml2, and is available in most Linux distros out-of-the-box. It's a command-line tool for doing XML validation against XSDs and DTDs, processing xpath expressions, and loads of other stuff. I mostly use it for the `--format` flag these days. If you do a lot of XML stuff, it's definitely a tool worth becoming familiar with (unless you're on Windows; I have no idea how people on Windows can get anything done without convenient tools like this :smiley_cat:)

hazendaz:
I'll have numerous things to test this on in coming days.  Thanks so much!

ctubbsii:
I released 2.9.0 with this.

hazendaz:
I've started testing with this on some rather large projects.  I've seen a couple of things that appear like the formatting failed and removed all spacing so tags are all against left side.  I wanted to mention this but otherwise need more time to try to understand what occurred before I open an issue.  Otherwise this support is hugely better than the original jsoup version and almost close enough to leave turned on (will require to work through a few things such as this first but super close).

jam01:
It's been pretty solid on a few small projects I've tried it on. If you want you can share a sample of what failed and we can create some tests off that.

PR #317, opened 2020-01-01
[ci] Format the xmlformatter file

hazendaz:


ctubbsii:
@hazendaz I don't think you need to create a PR for these kinds of things, unless you really want to. For something like this, I'd have just pushed directly to the repo. If you do do a PR, I think you could also squash it, rather than retain a merge commit, though I understand why you might not want to do that if you're issuing PRs from your own fork's master branch instead of a separate feature branch for the PR.

hazendaz:
I use github desktop.  The merge commit is extremely useful to know where I'm at.  A PR with squash option when more than one commit causes merge conflicts on rebasing which I heavily use.  With the 100+ repos I work on I typically do this on all for the clarity.  The only general time I might not is in the middle of a release and something is discovered.

ctubbsii:
No problem. Personally, I prefer feature branches in my fork that get squashed and deleted, rather than developing on my fork's master branch and dealing with merges. I think it keeps the history more easily readable. But do whatever works for you.

PR #325, opened 2020-03-28
Bring all libraries up to date, moderizer config change, remove deprecated compiler options

hazendaz:
- note added jackson annotations directly as they refuse to keep the underlying using that ever aligned release.  This ensures we have same version.  Seems silly on their part since they keep releasing it but they have discussed stopping that release and there was a lot of push back for this very reason.  Not ideal but not entirely bad to keep in sync on our part.

- jsdt-core update to 2.7.1 is not in this yet as I just released it.  I'll follow up with that in next few hours.  With stability of travis questionable when that occurs, want to hold until it's in all release channels.

- modernizer supports up to jdk 15 now.  However, it uses a legacy setup.  Since we require jdk 9 to build, the compiler options that are deprecated are not used so I removed them.  I adjusted this to use the release property.

- All other plugins used at this point that did share the compiler options have since received release that I'm currently aware of.  If you see any concerns there let me know.  I did same on jsdt-core and it had no issues with full release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,8 +117,6 @@
     <junit.version>5.6.0-M1</junit.version>
 
     <maven.compiler.release>8</maven.compiler.release>
-    <maven.compiler.source>1.8</maven.compiler.source>
-    <maven.compiler.target>1.8</maven.compiler.target>
 
     <!-- skip standard site deployment, because site is deployed using github plugin instead -->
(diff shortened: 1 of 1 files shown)


hazendaz:
Sent fix to go back to jsoup 1.12.1 until I can confirm test configuration (hashes need changed but need to do that together).
Added in skip on jackson annotatinos being directly added.

PR #358, opened 2020-06-21
[config] Sort the java.xml file so we can see what is actually being …

hazendaz:
…upgraded and/or missing

Eclipse does not keep this in order.  It does no harm in doing so.  This is as-is.


hazendaz:
I will be sorting all the files from eclipse.  I had noted this a number of times previously. We are extremely out of date and most use the defaults in the wild and hense a number of recent questions / asks because we are missing a lot.  In fact, we are 7 entire versions behind 2020-06.  This PR is as-is sorted.

PR #367, opened 2020-06-27
[cache] For skipped items, continue to write the cache and skip the w…

hazendaz:
…rite

add further commetns to help understand what is going on


hazendaz:
Alternative fix to #313

hazendaz:
An adjustment will be coming to this as the skip write to hash fails to pick up new value.

hazendaz:
Going to move forwards with this.  This works and has been tested numerous times now.

PR #372, opened 2020-06-27
[ci] Move cache file to basedir in folder .cache

hazendaz:
New logic to longer live the cache.  This allows continued builds to stop reformatting files if unnecessary.  The cache will live longer and as such it will be looked up first before any attempt to format at all times unless cache is deleted.  This speeds up initial processing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ nb-configuration.xml
 .idea/
 *.ipr
 *.iws
+
+### Formatter Cache ###
+.cache/
--- a/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -115,6 +115,9 @@ public class FormatterMojo extends AbstractMojo implements ConfigurationSource {
     @Parameter(defaultValue = ".", property = "project.basedir", readonly = true, required = true)
     private File basedir;
 
+    @Parameter(defaultValue = "${project.basedir}/.cache")
+    private File cachedir;
+
     /**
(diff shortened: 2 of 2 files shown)


ctubbsii:
This change causes problems... it creates a cache directory in every single downstream project that uses the plugin. This is a problem. The cache should be off by default, or tucked away inside the Maven `target/` directories. Alternatively, it should be disabled, unless the user explicitly sets a directory for it.

hazendaz:
Original is in target but loses ability to survive mvn clean.  Adding .cache to gift ignore solve it.

hazendaz:
btw - in case it wasn't clear, this is a parameter you can add to the build to control where it is cached if that default is not wanted.

  @Parameter(defaultValue = "${project.basedir}/.cache")
    private File cachedir;

So <cachedir>${project.build.directory}</cachedir> would restore prior behaviour.  To be far, I didn't run a test to confirm that though.  Ultimately I just felt more inclined that gitignore for most usage would suffice to allow mvn clean actions without blowing away the cache as that lost the value.  One might even want to check that in for faster CI system (assuming its same system line endings).

ctubbsii:
With the default the way it was before, I was satisfied with the difference between `mvn package` vs. `mvn clean package`, as being expected. I expect `clean` to blow everything away, so everything starts from scratch. That's why I use it!

I suspect the cache only affects a few people (most are probably not even aware there is a cache at all, at least not until 2.12), since it only really matters for big projects. However, having to do a .gitignore on the directory affects *everybody*.

In any case, we can continue this conversation on #382 and resolve it there.

hazendaz:
Correct. Feel free to set it back to target folder as default.  After seeing our monthly downloads, it would be months if we waited for more feedback.

PR #383, opened 2020-07-19
Add javadoc for cache change, use latest formatter on self, and add missing support for useSourceStringValue on CSS formatter

hazendaz:


hazendaz:
issue on travis is unrelated issue with pulling from sonatype.

PR #407, opened 2020-10-22
support multi-module projects

dinstone:
add multi-module projects config file support


ctubbsii (commented):
Hi @dinstone . Thank you for your PR.

However, this change doesn't appear to do anything useful. The user-specified path is already relative to the basedir of the module being executed, if it is specified as a relative path, and relative to the root of the filesystem if specified as an absolute path.

In addition, it looks like it might convert absolute paths to relative ones first, which is *very* surprising and confusing behavior.

Also, this looks like it would break the loading of resources from within jars by forcing the resource path to be always treated like a path on the filesystem (if it exists), rather than loading from a dependent resource jar.

The project already supports multi-module projects and absolute paths to config files. If you have questions about how to configure a multi-module project, feel free to ask, but I don't think these changes will help at all. Sorry.

PR #457, opened 2021-04-19
Add capacity to use a custom base diretory

brun-nico:
In some rare use case it could be good to use the plugin outside the project to format. For example when all the project don't share the same parent pom we could use a generic pom to automatically format code at pre commit like [this](https://github.com/brun-nico/formatter-maven-plugin).

To do that it's necessary to specify the base directory because in this case the base directory isn't equal to project.basedir.


ctubbsii (approved):
This seems reasonable to me.

PR #514, opened 2021-10-17
[actions] Move to java v2 and use zulu distribution

hazendaz:


ctubbsii:
Just to document the justification here (because I had questions about the justification and looked up the migration details):

* It's good to use the latest version of build tooling (to avoid bugs, and take advantage of improvements to make quality builds in test environment)
* The latest version of the `setup-java` GitHub Action requires the new `distribution` parameter
* The default distribution for `v1` was `zulu`, so keeping it that way for `v2` is the most minimal change (personally, I'd prefer the Adopt OpenJDK version, because that's what most people will probably be using in their dev environments)

See migration guide: https://github.com/actions/setup-java/blob/main/docs/switching-to-v2.md

hazendaz:
Adopt was replaced so that one ended life.  Eclipse has a different name now and it's a bit behind at least with jdk 18 early builds.  Adopt still works just doesn't get updates now.  I had been using Adopt too before someone informed me with a good link on what was happening.   I don't have that link handy at the moment.  I didn't know Zulu was really original so that's good.

ctubbsii:
Holy crap, this is a branding nightmare. The terminology is so unnecessarily confusing.

hazendaz:
Very much so!

PR #526, opened 2021-10-30
[css] Add patch to retain \9 during css formatting as cssparser conve…

### LINKED ISSUE #522: cssparser switches '\9' to tab
A bug exists in cssparser formatting which does not handle the formatting properly in regards to an IE 7, 8, and 9 hack.  We happen to have this hack in our test file but its converted to 'tab' and subsequent formatting of same file will entirely remove it.  Because these come in pairs and are easily verified, we need a patch to ensure we are not doing this wrong conversion.

https://stackoverflow.com/questions/8004765/css-9-in-width-property

In order to properly show this, our tests need to further change to run the first pass as is then run a second pass to confirm results are the same on repeated formatting.  This specifically as a test flushed this issue out and let to a patch.

hazendaz:
…rts to \t

This fixes #522


hazendaz:
will open new PR as missed updating the hashes.

PR #582, opened 2022-02-02
Got rest of trailing javadocs which shows why next release turns that on by default and one second pass if statement fix

hazendaz:


hazendaz:
This is last of code cleanup I'm performing via Eclipse.  I have ran 3 full passes after completion to just make sure and rechecked all parameters to make sure nothing missed.  I suspect some discovery after going back through code and understanding why some things changed and at least one case where full qualification causes formatting behaviour that is undesirable.  As I see those, I'll certainly make some adjustments for readabilty.

PR #588, opened 2022-02-02
Fix joup counting issue.

hazendaz:
Jsoup fails to count their indenting properly.  While they have a website to send test data for them to review they do not expose all the parameters.  So anything with 1 character indent is fine.  Its all the other indents that have issues.  The first commit here can be ran on its own still in single test mode then look at resulting second html page.  It will get extra 1 character on quite a number of lines.  Typically those are tag breaks with data on the second line.  This fix iteslf is extermely ugly but resolve the issue which I don't think jsoup is going to resolve.  They closed my original ticket stating fixed but I asked to reopen as not fixed.  They said they were looking but that was first week in November so nothing since and we are on the latest snapshot.  So like the others around jsoup fixing what they have failed to fix with hopes one day we remove this code.

### DIFF
--- a/src/main/java/net/revelc/code/formatter/jsoup/JsoupBasedFormatter.java
+++ b/src/main/java/net/revelc/code/formatter/jsoup/JsoupBasedFormatter.java
@@ -15,6 +15,8 @@ package net.revelc.code.formatter.jsoup;
 
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
 
@@ -38,6 +40,9 @@ public abstract class JsoupBasedFormatter extends AbstractCacheableFormatter imp
     /** The Constant REMOVE_TRAILING_PATTERN. */
     private static final Pattern REMOVE_TRAILING_PATTERN = Pattern.compile("\\p{Blank}+$", Pattern.MULTILINE);
 
+    /** The Constant RESET_LEADING_SPACES_PATTERN. */
+    private static final Pattern RESET_LEADING_SPACES_PATTERN = Pattern.compile("^\\s+");
+
     /** The formatter. */
     private OutputSettings formatter;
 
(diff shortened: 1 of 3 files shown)


src/main/java/net/revelc/code/formatter/jsoup/JsoupBasedFormatter.java
```diff
@@ -70,6 +75,30 @@ public String doFormat(final String code, final LineEnding ending) {
         // but our tests fail to do so thus we are duplicating this until jsoup fixes bug.
         formattedCode = REMOVE_TRAILING_PATTERN.matcher(formattedCode).replaceAll("");
 
+        // TODO: Fixing jsoup counting issue which occurs when more than one indentation (jsoup can have mixed line
```
hazendaz, line 78: I may need to document here more to be clear on what this is doing.

- If indent amount > 1 as only > 1 has the bug
- setup the current lineSize counter
- setup the new lineSize counter
- setup remainder
- Jsoup does not adhere to all same line endings.  So \r? as that happens on windows only in some areas based on their results.  So that along with \n or not.
- The resulting array is the entire thing that jsoup had just formatted
- Create an array to add elements into as we process the fix from the array
- loop through all lines of the array
- If the lineSize is 0, there is nothing to do as no indention on that line.
- Get the newLineSize by trimming all leading spaces
- If line size of original does not match that of new one, attempt to get the remainder by subtracting and dividing the result.
- If remainder is > 0, they are wrong on the count so just substring it properly
- In all cases in that loop either untouched or touched are added to the new array list
- Once completed, using String.join with the line endings we actually want against all lines.

The result here is that the first commit issue where they have many lines counted wrong, those are all corrected.

src/test/java/net/revelc/code/formatter/html/HTMLFormatterTest.java
```diff
@@ -36,12 +36,10 @@
     void testDoFormatFile() throws Exception {
         // FIXME Handle linux vs windows since this formatter does not accept line endings
         final var expectedHash = LineEnding.LF.isSystem()
-                ? "9774ef68195bb7bd9c1d0eaed6b3c7536633d0d916ad52187005ec2c89e0fcfe42e14dd4576453cb14abaef0352ef0b7a275c015e7cc124b8b4da481f423d030"
```
hazendaz, line 39: At this point, the two pass test is possible to work and was last in issue.  There is still one more bug after this that is more trivial that I have a PR out for but after 3 months I figure its going no where.

PR #592, opened 2022-02-02
Add sort and removal of timestamp from cache file for fuller automation / repeatability / First integration test

hazendaz:
- Sort and removal of timestamp do not appear to delay processing enough right now to warrant adding more if caching is being used.  Can follow up if we have more data to show that has any substantial impact.
- Sort and timestamp ensure that any unexpected write of file will not change the file if nothing changed.  This is prominant on maven plugins with help mojo as maven sets that in generated sources which is then picked up by formatting then promptly is moved by maven after plugin-plugin run.  This behaviour causes the current protections to see that formatting occurred and as such write the cache.  The cache itself is a properties file so it suffers from two major issues.  The first being the timestamp is added everytime it is written which is problematic at best and cannot be disabled (at least last confirmed at jdk 8, did not check to see if we have any new feature to skip that).  The second is that property file is just a hashtable and that is not deterministic on how it will even write the file.  Thus sorting it and removing the timestamp will ensure in such a case we don't have this issue.  It will further help if someone is saving the cache and needs to look at it since all will be ordered nicely.
- The first integration test and associated github action runs this against both jdk 11 and 17.  We should be doing same with general maven but this will cover that.  The users of library need to know this isn't going to break on higher jdks and at very least our lowest support.

### DIFF
--- a/.github/workflows/it.yaml
+++ b/.github/workflows/it.yaml
@@ -1,0 +1,43 @@
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


hazendaz:
Fixes #592

PR #595, opened 2022-02-10
support multiple-object json file

XenoAmess:
https://github.com/revelc/formatter-maven-plugin/issues/594

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -334,6 +334,12 @@
       <artifactId>junit-jupiter-engine</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <version>${junit.version}</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <!-- TODO: Allow pull from snapshots while on jsoup snapshot, remove once finalized -->
--- a/pom.xml
+++ b/pom.xml
@@ -337,7 +337,6 @@
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-params</artifactId>
-      <version>${junit.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/src/main/java/net/revelc/code/formatter/AbstractCacheableFormatter.java
+++ b/src/main/java/net/revelc/code/formatter/AbstractCacheableFormatter.java
@@ -38,7 +38,7 @@ public abstract class AbstractCacheableFormatter {
     protected Charset encoding;
 
     /**
-     * Inits the.
+     * Inits the AbstractCacheableFormatter.
      *
      * @param options
      *            the options
@@ -129,6 +129,9 @@ public abstract class AbstractCacheableFormatter {
 
     /**
      * Do format.
+     *
+     * notice that when calling this function, {@code ending} here MUST equals lineending in config when
+     * {@link #init(Map, ConfigurationSource)} init.
      *
      * @param code
      *            the code
--- a/src/main/java/net/revelc/code/formatter/json/JsonFormatter.java
+++ b/src/main/java/net/revelc/code/formatter/json/JsonFormatter.java
@@ -87,7 +87,7 @@ public class JsonFormatter extends AbstractCacheableFormatter implements Formatt
         return doFormatWithMultipleJsonFileForbidden(code, ending);
     }
 
-    protected String doFormatWithMultipleJsonFileForbidden(final String code, final LineEnding ending)
+    private String doFormatWithMultipleJsonFileForbidden(final String code, final LineEnding ending)
             throws IOException {
         // note: line ending set in init for this usecase
         final var json = this.formatter.readValue(code, Object.class);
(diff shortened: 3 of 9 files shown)


XenoAmess:
> Assuming JSON objects are separated by newlines is not a great way to read separate JSON objects, since JSON itself is not sensitive whitespace.

Yes we do not use newlines to seerate it. you can see the codes for detail.

EDIT: also, these tests fail

I see. investigating.

XenoAmess:
@ctubbsii
seems the original codes have some bug already.
will fix it also in this pr.

XenoAmess:
Ah!
I see why.
Seems I use it wrongly.
The formatter must be called doFormat with exactly same line-end config with when it initialized.

XenoAmess:
So what I need to do is rewrite the tests.

XenoAmess:
done.
github actions passed in forked repo.

ctubbsii (changes requested):
I reviewed this. It seems that formatting the multiple object with the multiple objects forbidden, just results in output that preserves only the first object. Everything else gets truncated silently, no error. That's not quite what I was expecting. I was expecting the code to throw an exception that the entire object was invalid.

If it's not going to throw an exception in the case of multiple values when that flag is off, I agree with your original assessment that it should probably just format all objects, regardless of how many it finds. It's better to format invalid JSON than to delete user's data.

Also, please be mindful of creating methods more visible than they need to be. The two new `doFormatWithMultiple*` methods that you created are marked `protected` and result in new javadoc warnings due to missing javadocs. They can be `private` instead. Also, please run `mvn package` to format any import statements and code in accordance with this project's standards.

XenoAmess:
> It seems that formatting the multiple object with the multiple objects forbidden, just results in output that preserves only the first object. Everything else gets truncated silently, no error.

Yes and that is what the original codes before this pr do.

> It's better to format invalid JSON than to delete user's data.

Yes, I think so.

> Also, please be mindful of creating methods more visible than they need to be. The two new `doFormatWithMultiple*` methods that you created are marked `protected` and result in new javadoc warnings due to missing javadocs. They can be `private` instead. Also, please run `mvn package` to format any import statements and code in accordance with this project's standards.

Got it. Will check it 15 hours later when get off work.

XenoAmess:
@ctubbsii Hi. problems fixed. please review again. thanks.

hazendaz:
LGTM!  Getting ready to release, merging this.

hazendaz:
@ctubbsii Will hang a few hours on release but appears all was done here.  If anything else was missed here or needs adjusted let me know and I can take care of it before I push release.  We will either way be releasing in coming weeks again anyways.  So this is really to shore up 2021-12 and then just hopefully a quick release to 2022-03.

ctubbsii:
@hazendaz This wasn't quite ready. There was definitely changes that needed to be made. The author said they fixed the issues I raised, but I don't think they did. I haven't had time to make a comment, but I think this isn't quite done.

ctubbsii:
@hazendaz In general, please don't merge something that is being reviewed by somebody else, unless you can either confirm that they have completed their review, or are unresponsive. A simple ping would have revealed that I didn't think this was ready to merge. You jump the gun too much, and take too many aggressive actions unilaterally. It's becoming frustrating. Please slow down, and try to collaborate more, please. Thanks.

ctubbsii:
@hazendaz wrote:
> Will hang a few hours on release but appears all was done here

Please don't. A "few hours" is not enough time for me to address the changes I'd like to see in this PR. Most open source communities operate on more friendly "days" timelines, not "hours". It's my weekend right now, and my time off to relax. If you'd like, I can fix the issues with this PR that I have, and either you or I can do a release tomorrow. But, please don't release it today, when I have outstanding issues with this PR after merging it without checking with me.

hazendaz:
That is fine, what it else needs done to it?  Maybe default to false but rest is very clean and net new code so I'm not seeing what wasn't done.

As to release we haven't released since September releases and we are really close to the next eclipse cut and 2 months past dates set for this one in milestones.  I can wait until tomorrow no worries but need this soon.  It's holding up release train at my day job and is vulnerable on top of it with current releases.

XenoAmess:
Sorry if I caused any trouble.
If there be anything in the codes need me to change / fix, I'm glad to help.

ctubbsii:
@XenoAmess wrote:
> Sorry if I caused any trouble. If there be anything in the codes need me to change / fix, I'm glad to help.

It's not much trouble. I think the main follow-up was that my review comment in https://github.com/revelc/formatter-maven-plugin/pull/595#pullrequestreview-883799302 was that I was agreeing with you that we don't actually need the boolean and the two separate code paths, and that it made more sense to just format all detected objects. This implies the code can be further simplified beyond what was merged. I can perform the cleanup as a subsequent commit.

hazendaz:
@XenoAmess no worries, assumed this was for some other use case.  The one I pushed has what you need so it should work better for your needs.  But what a difference 24 hours makes.  Guava released a new version over night :(

hazendaz:
sorry ignore my comment, wrong PR / repo came up. deleted it.

ctubbsii:
Completed my cleanup in 078173e2dcf24124a4a2d7c273ab71ecd508e121

PR #596, opened 2022-02-10
[tests] Add a test for showing bug in json formatter.

XenoAmess:
Hi.

This is just for showing the bug we found.

image

although we use LineEnding.LF, it still contains CRLF in the output when we run it on windows.


XenoAmess:
Ah!
I see why.
Seems I use it wrongly.

PR #598, opened 2022-02-27
Move to fork of jsoup until jsoup team releases code

hazendaz:
allows us to release again.  Code line is the same outside of some minor pull requests not yet merged to jsoup that are used for the overall build.


hazendaz:
this is ok, was just too fast on push, confirmed on couple of these, merging now.  Will be back later to prepare for the final release.

PR #642, opened 2022-07-08
[pom] Move back to upstream jsoup 1.15.2

hazendaz:
- patches applied
- fixes additional spacing issue
- fixes duplication on license header

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -266,11 +266,6 @@
       <artifactId>jackson-databind</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


hazendaz:
Upstream jsoup now has variations of initial patches I had sent that resulted in us running on a fork.  It further had a strange spacing issue that was fixed and an issue with duplicate license header we had on that html page (had been unnoticed).  In real life usage, I had not seen those two specific issues but its better and why the hash changed.

Issue #270, opened 2018-02-16
validate does not check all the files at once


victornoel:
When I run the `validate` goal, it fails at the first problematic file instead of showing me all the files that fail.

If the objective is to use it in the CI for example, this is very limiting to only show one error…

hazendaz:
Not really.  It simply means formatting was not run.  So when CI fails, developer should run a proper build and commit the result which may be many files.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

victornoel:
I understand that, but it would be nicer if we could get all at once: even locally, it is tedious to run `validate` one file per one file to correct each of them.

It is not always possible to use eclipse's formatting on a whole project, for example if one configure some exclusions from the validation.

PR #644, opened 2022-07-16
[draft] Solution for issue 270 to show all validation errors

hazendaz:
Initial attempt at fixing #270.  This is draft to gather input for feedback of supporting this feature.  It does not have an integration test but steps to test are below.

To test
- Checkout branch
- Run build to generate local snapshot for test usage
- Add extra spacing to 2 or more files
- Run 'mvn net.revelc.code.formatter:formatter-maven-plugin:2.20.0-SNAPSHOT:validate' which will show new behavior of split from output of first file detected that violates formatting rules.
- Run `mvn net.revelc.code.formatter:formatter-maven-plugin:2.20.0-SNAPSHOT:validate -D"formatter.failOnFirstValidationError=false"` which will show multiple files in violation.

### DIFF
--- a/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -733,8 +733,11 @@ public class FormatterMojo extends AbstractMojo implements ConfigurationSource {
             return;
         }
 
-        // Now write the file
-        if (!dryRun) {
+        // Now write the file or log during dry run
+        if (dryRun) {
+            String errorMessage = String.format("File '%s' has not been previously formatted.", file);
+            this.getLog().error(errorMessage);
+        } else {
             this.writeStringToFile(formattedCode, file);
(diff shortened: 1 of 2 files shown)


ctubbsii (commented):
The approach seems reasonable. I like that it defaults to fast failure, which is the current behavior. It might be nice if the batch report of failures could be saved in a file in the target directory.

hazendaz:
Thanks, thinking this would come post 2022-06 release with target around 2022-09 given still need to build out IT tests and batch report file consideration.

ctubbsii:
@hazendaz 2022-09 is out. Wasn't sure if you had time to revisit this, so we could include it. No rush, though. I figure we could wrap up 2.21 for October for 2022-09.

hazendaz:
I say just go ahead and release 2022-09.  I had not come back to this yet and currently no power in tampa area day 3 since Ian.  We are ok though.

PR #645, opened 2022-07-17
Jsoup follow-up formatting patches

hazendaz:
- Use XXX rather than TODO as entries are not really todo's but need called out if upstream fixes
- Pull up line ending normalization as jsoup will cause non normalized data during its process even if file was previously normalized.  This was mixed into another fix.
- Add clarifications within index counter fix.  Issue typically shows up on span, div, a, possibly others where it still adds one extra space in violations to what was asked for index size
- Removed header line break fix as that was fixed in the fork we used and now also fixed in jsoup directly
- Add new line to end of the file

The existing checks were all verified to still be considerations that have not yet been fixed in jsoup pretty print.  There are 4 patches as follows:
- Removal of trailing space on each line
- Normalization of line endings (jsoup processing internally is LF which will cause mixed results in windows)
- Index counter problem resulting in some lines getting 1 extra space typically
- Add new line to end of file if missing

### DIFF
--- a/src/main/java/net/revelc/code/formatter/jsoup/JsoupBasedFormatter.java
+++ b/src/main/java/net/revelc/code/formatter/jsoup/JsoupBasedFormatter.java
@@ -77,18 +77,20 @@ public abstract class JsoupBasedFormatter extends AbstractCacheableFormatter imp
 
         // XXX: Fixing jsoup counter issue when more than one character indentation until jsoup fixes bug.
         if (this.formatter.indentAmount() > 1) {
-            int lineSize;
-            int newLineSize;
+            int lineLength;
+            int trimLineLength;
             int remainder;
             // jsoup can have mixed line ending content
             String[] lines = formattedCode.split("\\r?\\n");
             List<String> newLines = new ArrayList<>(lines.length);
             for (String line : lines) {
-                lineSize = line.length();
-                if (lineSize != 0) {
-                    newLineSize = RESET_LEADING_SPACES_PATTERN.matcher(line).replaceAll("").length();
-                    if (lineSize != newLineSize) {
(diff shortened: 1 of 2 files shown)


src/main/java/net/revelc/code/formatter/jsoup/JsoupBasedFormatter.java
```diff
@@ -71,36 +71,45 @@ public String doFormat(String code, final LineEnding ending) {
...
+
+        // XXX: Fixing jsoup counter issue when more than one character indentation until jsoup fixes bug.
         if (this.formatter.indentAmount() > 1) {
-            int lineSize;
-            int newLineSize;
+            int lineLength;
+            int trimLineLength;
             int remainder;
-            String[] lines = formattedCode.split("\\r?\\n");
+            lines = formattedCode.split(ending.getChars());
```
ctubbsii, line 87: This line's change will leave behind extra chars if the detected line endings aren't consistent throughout the file (like if there's a mix of line ending styles in the file). The current code will split on any line, regardless of the line ending style.

hazendaz:
The splitter on mixed content was moved above this section so it only needs to look at mixed content the one time.  The subsequent ones have line ending then normalized.  See line 80.

src/main/java/net/revelc/code/formatter/jsoup/JsoupBasedFormatter.java
hazendaz: Added additional commits to help walk through changes on jsoup post pretty print processing.  Specific issue was done earlier in process to get it out of mixed patch so that each patch stands on its own.

hazendaz:
Going to go ahead and merge as concern addressed by adding more clarity to the jsoup patches and our build is fully testing it with 2 passes on file.

PR #673, opened 2022-12-06
[672] Add a parameter for including resources when formatting.

### LINKED ISSUE #672: Allow resources to be included when formatting
**Description**
Resources can be included by overriding the `<directories/>` or the `<sourceDirectory/>` and `<testSourceDirectory/>` configuration parameters. However, this requires that the user redefines each resource directory. It would be nice to have a new parameter, `<includeResources/>`, which would include the `MavenProject.getResources()` and `MavenProject.getTestResources()`.

**Describe the solution you'd like**
A configuration property that can be defined to allow for resources to be formatted.

**Describe alternatives you've considered**
Using the various options to override directories. However, since resources are an array it makes it more verbose and possibly more overrides in multi-module projects.

**Additional context**
No additional context at this time.

jamezp:
resolves #672

Signed-off-by: James R. Perkins <jperkins@redhat.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -327,6 +327,12 @@
     </dependency>
     <dependency>
       <groupId>org.apache.maven</groupId>
+      <artifactId>maven-model</artifactId>
+      <version>3.8.6</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
       <version>3.8.6</version>
       <scope>provided</scope>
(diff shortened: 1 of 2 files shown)


hazendaz:
As this stands, its a nice improvement.  We may want to look at the older way of doing things to avoid collisions in case intent was global override ignoring others but given its flag based its fine.  In fact, this is something I see needed on another plugin I work with (htmlcompressor-maven-plugin) for same reasons.  Manually trying to capture actual path if were to be done and array based is a guessing game.  Anyways, merging this as a good improvement and may toy around a bit afterwards with it when I get some time.

jamezp:
Excellent, thank you!

PR #683, opened 2023-01-16
Align usage better with maven slf4j usage, bump few versions

hazendaz:
commons-logging as independent library is long obsolete and should not come into our stack.  Given maven is slf4j based, use jcl-over-slf4j as a direct replacement so that logging, if any, actually is picked up from that transient usage.

Update few other dependencies recently released.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -292,6 +292,12 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-digester3</artifactId>
       <version>3.2</version>
+      <exclusions>
+        <exclusion>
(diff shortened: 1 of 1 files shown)


hazendaz:
on commons-logging, I have a separate effort that scans and blocks bad logging, it picked up formatter plugin as being exposed, so patched.  I'd expect if a new commons digester is ever released it would do away with that library.  The one currently used here and last released is currently 11 years old.  Usage of jcl-over-slf4j gains some benefit here in that any lost logs would actually get logged now.

PR #697, opened 2023-03-01
Add missing default items to the json properties file and bump maven plugin plugin to 3.8.1

hazendaz:


hazendaz:
Not sure when those were added but we missed adding them to the file.  They are set to the defaults.  Caught that on a side email someone sent on wanting to see comments introduced which I also wrote but not sure I wish to submit as they never responded back after asking some questions but this here was worth making sure correct.  Library update is really a downside of dependabot running on forks so applied that in too.

PR #708, opened 2023-03-09
Enforce mismatch with prerequisites

matteobaccan:
Hi

I have tried to rebuild the project, but there is a mismatch between prerequisites (maven 3..3.9) and enforce (maven 3.5.0) 
I have updated the prerequisites to 5.5.0

ciao
matteo

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
   </developers>
 
   <prerequisites>
-    <maven>3.3.9</maven>
+    <maven>3.5.0</maven>
   </prerequisites>
 
   <scm>


hazendaz:
Forget offhand if we did this...a gh-actions using maven 3.3.9 to confirm would be useful.  Typically speaking most of maven is only enforcing 3.2.5.  And of course there won't be legacy patches of maven and plugin does require java 11.  So raising it does have some rational reason, but we need to make sure it's true with gh action.  If we don't have one already I'll build one.  Last thing I think we want to do is state we require higher version then get issues stating we don't.

matteobaccan:
Hi

I'm not sure that it's Netbeans bugs, or better, if you declare the minimum version 3.3.9 and using a standard "clean install" with 3.3.9 version of maven the process is blocked for an enforce, I ask me why, if the minimal declared is 3.3.9.
Now I understand that build is made with 3.5.0 and the run is 3.3.9, but after the first compile, I haven't understood this, and I have made the modification for this reason.

ciao
matteo

ctubbsii:
Based on the above discussion, I'm going to close this PR for now.

hazendaz:
@ctubbsii Setting up GH-Action for whatever our true lowest support is including matching jdk run of usage is rather simple.  I've got similar setup over at spotbugs-maven-plugin [here](https://github.com/spotbugs/spotbugs-maven-plugin/blob/spotbugs/.github/workflows/it-maven339.yaml).  This does show why maven wrapper is important as its necessary to pull that off.  If willing to allow maven wrapper back in, I can work this up.  Preference is to keep it uncoupled from the main run of checks as its more clear for the intent but it could be also just more steps to the maven.yaml that exists now.

hazendaz:
actually may not need to add wrapper directly, that example actually is calling out maven to switch so never mind on that.  Still the idea there is rather simple, just cause our IT tests to run on older copy.  May require a few tweaks but simple enough.

hazendaz:
Part of the issue faced here is a problem with building this plugin with 3.3.9 as indicated.  That is because of impsort that is flagged to require maven 3.5.0.

Failed to execute goal net.revelc.code:impsort-maven-plugin:1.8.0:sort (sort-imports) on project formatter-maven-plugin: The plugin net.revelc.code:impsort-maven-plugin:1.8.0 requires Maven version 3.5.0 -> [Help 1]

Of course that is provider side so its meaningless to consumer side, however simply pointing that out.

hazendaz:
Another update here, I was able to set impsort down and it still failed, turns out we actually required 3.5.0.  See https://github.com/hazendaz/formatter-maven-plugin/actions/runs/4507639224/jobs/7935599706 to see the actual requirement which was internal.

hazendaz:
And reality is at the moment, due to how we are structured, we cannot properly separate the producer from consumer to properly confirm.  The hang up is on our producer end.  So its probably right 3.3.9 works to run it but best we can do without a lot of restructure is to say 3.5.0 which @ctubbsii moved the code to and called it a day.  I'll keep an active test at that level to show at least that works but don't intend to restructure enough to see a true simulated toolchain usage at the current time.  Restructure would require independent setup of invoker to run the IT tests similar to what exists with license maven plugin.  This think for now this is good enough to just say we are 3.5.0 as the effort to show more simply isn't worth it as honestly who uses 3.3.9 still.

PR #711, opened 2023-03-13
[github] Delete dependabot - stop spamming forks

hazendaz:
Go manual entirely dropping dependabot so it stops spamming forks.


ctubbsii (commented):
I don't think this is necessary because of https://github.blog/changelog/2022-11-07-dependabot-pull-requests-off-by-default-for-forks/

I checked my fork, because I don't get these spam PRs at all (I've never gotten any PR from dependabot for my fork of this repo, actually), and it shows that dependabot is disabled when I go to my fork's settings at https://github.com/ctubbsii/formatter-maven-plugin/network/updates

PR #731, opened 2023-04-06
Refactoring

Kovarthanan-murugan:
I am new to software development and creating pull requests as it is mandatory for my college assignment. It would be helpful if pr is closed same as my previous pull request. 

1. In class FormatterMojo
    In method Execute
    Refactored complex method 
    created fileEncoding, formatFileDirectory, getFilesToFormat, formatFiles methods to reduce the complexity of the execute 
    method

2. In class FormatterMojo
   In method Execute
   Moved the method printDuration to formatterMojo class
   As this is the only place this method is used and close to the calling method

3. The class HTML formatter  implements jsoup class
    Pull done methods from jsoup class to HTML class.

4. In class FormatterMojo
    The readFileAsString and writeStringToFile methods used to do IO operation is extracted and kept 
    In new package file operation and in class fileIO
    To follow the single responsibility principle


hazendaz:
@Kovarthanan-murugan This is a lot to take at once.  Typically you would want to isolate smaller changes so its easier to review.  Some of this looks good but other parts I don't see how it makes it any better.  If you could focus on smaller pull requests with single items to address it would be easier to not only get reviewed but get it merged.  One thing I'd start with just because its super simple one would be the 'stringIncrementor' as its simple and makes sense and likely you used something like sonar or checkstyle that flagged that.

I would not bother with #2.  That is separate on purpose as it has zero to do with the main class.  Its a utility that yes that is the only class that uses it but having more classes smaller in scope is better design.  

On #3, I would have to look but jsoup is used by more than one spot so moving those out would lose what its doing as all the files are treated the same for that formatter.  Going off memory here so again would have to actually look to confirm.  The extensions only do small portion so we need that to stay where it is.  If anything it was just hiding some details of jsoup as sort of out of scope for how this repo typically works for the other formatters.

On #4, most of that code would likely go away.  Its very legacy in nature and we require jdk 11 now so it may be better to use newer java IO routines which would lesson the code and not result in needing a utilities class for the same.  That is certainly something on our radar so something you could clearly excel at getting a pull request through :)

At any rate, good effort for your class exercise.  If you can come in with smaller items in general its always more likely to get it merged in.  Large changes like this make it really difficult to review even if 100% valid.  Try given a few small pull requests focused on one thing done well and lets see how much better that works out.

hazendaz:
as-is closing for now as too much going on, also look to making a better title so its a bit more clear than just large refactoring effort.  I think per other PR you had opened it was basic same response.  Unless it has some tangible value making the code either far easier to read, improves our processing, etc, then most likely it won't get picked up unfortunately.

hazendaz:
one thing you might do is look through our 'issues', see if there is anything within that you could take on and champion.  Not sure we have a lot for first timers in there but probably something that might be easy and we have just not spent the time to address.  thanks.

Kovarthanan-murugan:
> @Kovarthanan-murugan This is a lot to take at once. Typically you would want to isolate smaller changes so its easier to review. Some of this looks good but other parts I don't see how it makes it any better. If you could focus on smaller pull requests with single items to address it would be easier to not only get reviewed but get it merged. One thing I'd start with just because its super simple one would be the 'stringIncrementor' as its simple and makes sense and likely you used something like sonar or checkstyle that flagged that.
> 
> I would not bother with #2. That is separate on purpose as it has zero to do with the main class. Its a utility that yes that is the only class that uses it but having more classes smaller in scope is better design.
> 
> On #3, I would have to look but jsoup is used by more than one spot so moving those out would lose what its doing as all the files are treated the same for that formatter. Going off memory here so again would have to actually look to confirm. The extensions only do small portion so we need that to stay where it is. If anything it was just hiding some details of jsoup as sort of out of scope for how this repo typically works for the other formatters.
> 
> On #4, most of that code would likely go away. Its very legacy in nature and we require jdk 11 now so it may be better to use newer java IO routines which would lesson the code and not result in needing a utilities class for the same. That is certainly something on our radar so something you could clearly excel at getting a pull request through :)
> 
> At any rate, good effort for your class exercise. If you can come in with smaller items in general its always more likely to get it merged in. Large changes like this make it really difficult to review even if 100% valid. Try given a few small pull requests focused on one thing done well and lets see how much better that works out.



Thank you @hazendaz  for the feedback, it will really help me to do better.

PR #812, opened 2023-12-09
 Update for Eclipse 2023-12

ArloL:
I hope I did everything correctly ☺️

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -149,97 +149,97 @@
         <!-- transitive dependency of org.eclipse.jdt.core resolved here to avoid version range -->
         <groupId>org.eclipse.jdt</groupId>
         <artifactId>ecj</artifactId>
-        <version>3.33.0</version>
+        <version>3.36.0</version>
       </dependency>
       <dependency>
         <!-- transitive dependency of org.eclipse.jdt.core resolved here to avoid version range -->
         <groupId>org.eclipse.platform</groupId>
         <artifactId>org.eclipse.compare.core</artifactId>
-        <version>3.7.100</version>
+        <version>3.8.300</version>
       </dependency>
       <dependency>
         <!-- transitive dependency of org.eclipse.jdt.core resolved here to avoid version range -->
         <groupId>org.eclipse.platform</groupId>
         <artifactId>org.eclipse.core.commands</artifactId>
-        <version>3.10.300</version>
+        <version>3.11.100</version>
       </dependency>
       <dependency>
         <!-- transitive dependency of org.eclipse.jdt.core resolved here to avoid version range -->
         <groupId>org.eclipse.platform</groupId>
         <artifactId>org.eclipse.core.contenttype</artifactId>
-        <version>3.8.200</version>
+        <version>3.9.200</version>
       </dependency>
       <dependency>
         <!-- transitive dependency of org.eclipse.jdt.core resolved here to avoid version range -->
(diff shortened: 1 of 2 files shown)


hazendaz:
Seems ok to me but does mean we are raising jdk requirements.  Please do update the requirement here as this won't work with jdk 11.  The internals already state as much.  If I recall we didn't even bother releasing the last one as a result.  Plus there were issues with static imports with mockito 'when' that were entirely broken.  The caching is also still broken but that isn't relevant to this upgrade.  With the community as a whole mainly starting with spring forcing java 17 on everyone and now jakarta forcing at least 21 and tomcat, either accidentally or on purpose forcing jdk 22, we have no option but to drop old users.

src/site/markdown/eclipse-versions.md.vm
```diff
@@ -26,6 +26,7 @@ their compatibility with major Eclipse releases.
 
 formatter-maven-plugin | Eclipse<br/>Name | Eclipse<br/>Version | JDT Core | JDK<br/>Version
 -----------------------|------------------|---------------------|----------|----------------
+2.24                   | 2023-12          | 4.30                | 3.36     | 11
```
hazendaz, line 29: Make sure this says 17 for jdk requirement.

pom.xml
```diff
@@ -325,7 +325,7 @@
       <!-- bump this and run src/build/find-transitive-eclipse-updates.sh to update eclipse deps -->
       <groupId>org.eclipse.jdt</groupId>
       <artifactId>org.eclipse.jdt.core</artifactId>
-      <version>3.33.0</version>
+      <version>3.36.0</version>
```
hazendaz, line 328: Yah , mockito 'when' static usage is now fixed.

src/site/markdown/eclipse-versions.md.vm
ArloL: I just updated the branch.

hazendaz:
Just a note to add, from my fork which is using enforcer plugin separately to show the jdk requirement.

```
[INFO] Restricted to JDK 11 yet org.eclipse.jdt:ecj:jar:3.35.0:compile contains org/eclipse/jdt/internal/compiler/lookup/ParameterizedRecordComponentBinding.class targeted to JDK 17
[INFO] Restricted to JDK 11 yet org.eclipse.jdt:org.eclipse.jdt.core:jar:3.36.0:compile contains org/eclipse/jdt/internal/codeassist/complete/CompletionOnMessageSendName.class targeted to JDK 17
```

PR is good now, will merge.

ctubbsii:
I think there was something missing here. I added it in #826 . I also updated the parent POM that I've been working on to help clean up and simplify the maintenance of this and other projects in revelc (especially to get dependabot spam under control). I added a profile to do the byte code checking in version 6 of the parent POM (https://github.com/revelc/revelc/commit/f178fe8b88442e5ce8e8f5e411dc7232c752676f), but version 6 isn't released yet. I did test it with these changes, and with #826 to confirm that #826 closes the gap and makes JDK 17 a strict requirement.

ctubbsii:
Re: dropping old users. If there's a bugfix needed for 2.23, we can still patch it for users on JDK 11. However, it's likely that users will just stick with the old version as-is until they are ready to jump to JDK 17 (at least for builds, if not for runtime).

hazendaz:
Sounds good!

PR #834, opened 2024-03-09
[pom] Add missing relativePath set to nothing

hazendaz:
Parent poms that don't exist up one level, they need to not look up one level but instead look to m2 cache immediately.  In my case I have a pom level one up that is for cleaning my workspace and maven will issue configuration warnings stating this must be added and its best practice when using a parent pom that is not in same project.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,6 +20,7 @@
     <groupId>net.revelc.code</groupId>
     <artifactId>revelc</artifactId>
     <version>5</version>
+    <relativePath />
   </parent>
   <groupId>net.revelc.code.formatter</groupId>
   <artifactId>formatter-maven-plugin</artifactId>


hazendaz:
To show the issue.

```
[WARNING]
[WARNING] Some problems were encountered while building the effective model for net.revelc.code.formatter:formatter-maven-plugin:maven-plugin:2.24.0-SNAPSHOT
[WARNING] 'parent.relativePath' of POM net.revelc.code.formatter:formatter-maven-plugin:2.24.0-SNAPSHOT (...\formatter-maven-plugin\pom.xml) points at com.github.hazendaz:clean-target instead of net.revelc.code:revelc, please verify your project structure @ line 19, column 11
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
```

As good practice to make sure relativePath is set so it doesn't do the typical look up one directly but goes to .m2.  Same issue is on all the parent changes made to these projects.

PR #835, opened 2024-03-10
Update libraries and fix hash on tests for jsoup/jackson updates

hazendaz:
replacement of #831


hazendaz:
For jsoup, they changed a casing on one element to standard usage.  For jackson, I didn't see what exactly changed.  However, both needed the hashes updated to proceed.

PR #838, opened 2024-03-10
[fix] Add missing check for json to config checks

hazendaz:
Error notes all 6 formatters but only 5 were being checked.  Corrected so all 6 are checked.


hazendaz:
Minor issue as noted was missing json check but error indicated it was checking all including json.

PR #841, opened 2024-03-16
Make the json options private and move the object mapper creation closer to usage

hazendaz:


hazendaz:
On #836, I initially was making those options public but then added proper getter but missed setting one back private.  Also on another change I'm looking at noticed the object mapper is created pretty early so fixed that to create only at point it was needed.

PR #842, opened 2024-03-17
[GHA] Change caching to built in from setup-java

hazendaz:
See https://github.com/actions/setup-java?tab=readme-ov-file#caching-packages-dependencies as this is built in now.


hazendaz:
Ran jobs twice as cache is different than before.  Second run did use the new cache.  I presume the old just will fall off after time.

PR #843, opened 2024-03-17
[GHA] Add workflow_dispatch and let jobs run on any branch

hazendaz:


hazendaz:
There is no reason to lock GHA to 'main' branch.  This repo has 2 branches, main and gh-pages.  gh-pages does not include any of this so not applicable.  PRs to main are already covered as it is so dependabot is not affected by the change.  The point to having this is for those that contribute that have many branches with many different things such as myself or if we need to run extra branches here.  Its not the best to just block those because that means most that don't know how to make them run will send PRs that just fail, spam emails, etc.  This ensures that is not the case.  It also ensures users don't have to add special GHA changes to every branch they have as I have to do now if I want them to execute.  And finally, not everyone uses 'main'.  Therefore its about flexibility and not enforcing contributors to act in the way we want.  That never scales.

hazendaz:
For dispatch, see top of this https://github.com/revelc/formatter-maven-plugin/actions/workflows/maven.yaml, new button there now.  Its helpful if we just need to rerun the build for whatever reason.

PR #845, opened 2024-03-17
[actions] Add sonatype to auto deploy to snapshots

hazendaz:


hazendaz:
For this to work, it needs .mvn folder to be officially added which is part of maven wrapper.  It needs few secrets added as well.  Then snapshots will be available in sonatype.  For now only proposing this be done so we can have folks more easily test without having to build themselves.

hazendaz:
added the maven file.  All that is needed is to add secrets.

hazendaz:
Added necessary secrets, merged, confirming run now.

PR #896, opened 2024-06-03
[894] Use the formattedCode for the formattedHash.

### LINKED ISSUE #894: Formatter no longer formats in version 2.24.0
With version 2.23.0 of the plugin if I had the code `var result =  function.apply(input);` the formatter would correct the double space based on what I believe was this eclipse rule
```xml
<setting id="org.eclipse.jdt.core.formatter.insert_space_after_assignment_operator" value="insert"/>
``` 
Then when the checkstyle plugin executed further in the lifecycle the rule
https://checkstyle.sourceforge.io/checks/whitespace/singlespaceseparator.html
would pass.

When we migrated to Java 21 I had to disable the formatter, and also checkstyle.
With version 2.24.0 of the formatter plugin I reactivated both of these plugins, and now checkstyle is catching the double space.
Why is the formatter not formatting the code? I have confirmed it is being invoked.
Do I need to update the custom eclipse formatting xml?

I checked the plugin documentation (https://code.revelc.net/formatter-maven-plugin/examples.html#setting-compiler-version which I see is out of date, as is the homepage readme https://github.com/revelc/formatter-maven-plugin/blob/main/README.md)
I added the last three configurations items just in case, but it makes no difference
```xml
<configuration>
            <encoding>UTF-8</encoding>
            <configFile>eclipse/formatter.xml</configFile>
            <skipCssFormatting>true</skipCssFormatting>
            <skipFormatting>false</skipFormatting>
            <skipHtmlFormatting>true</skipHtmlFormatting>
            <skipJavaFormatting>false</skipJavaFormatting>
            <skipJsFormatting>true</skipJsFormatting>
            <skipJsonFormatting>true</skipJsonFormatting>
            <skipXmlFormatting>true</skipXmlFormatting>
            <compilerSource>${maven.compiler.release}</compilerSource>
            <compilerCompliance>${maven.compiler.release}</compilerCompliance>
            <compilerTargetPlatform>${maven.compiler.release}</compilerTargetPlatform>
          </configuration>
```

jamezp:
resolves #894

### DIFF
--- a/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -806,19 +806,19 @@ public class FormatterMojo extends AbstractMojo implements ConfigurationSource {
             formattedHash = originalHash;
             // Default to formatted hashing for unknown type otherwise include formatter options
         } else if (file.getName().endsWith(".java")) {
-            formattedHash = this.sha512hash(originalCode + this.javaFormatter.getOptions().hashCode());
+            formattedHash = this.sha512hash(formattedCode + this.javaFormatter.getOptions().hashCode());
         } else if (file.getName().endsWith(".js")) {
-            formattedHash = this.sha512hash(originalCode + this.jsFormatter.getOptions().hashCode());
+            formattedHash = this.sha512hash(formattedCode + this.jsFormatter.getOptions().hashCode());
         } else if (file.getName().endsWith(".html")) {
-            formattedHash = this.sha512hash(originalCode + this.htmlFormatter.getOptions().hashCode());
+            formattedHash = this.sha512hash(formattedCode + this.htmlFormatter.getOptions().hashCode());
         } else if (file.getName().endsWith(".xml")) {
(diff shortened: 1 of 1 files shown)


src/main/java/net/revelc/code/formatter/FormatterMojo.java
```diff
@@ -792,7 +792,7 @@ protected void doFormatFile(final File file, final ResultCollector rc, final Pro
 
         // Process the result type
         if (Result.SKIPPED.equals(result)) {
-            rc.unchangedCount++;
+            rc.skippedCount++;
```
jamezp, line 795: I wasn't sure if this line should be changed or not. I changed it for testing and forgot to revert it before committing. If this change should be removed, please let me know.

hazendaz: Change it back for purpose of this pull request.  I do want to look more at this part after as it's still confusing on naming.

jamezp: Updated. Sorry about that, I should have just done it last night :)

Issue #929, opened 2024-07-30
JSON formatting option "spaceBeforeSeparator" is ignored


dmak:
**To Reproduce**

Run `mvn formatter:format` on the attached project [test.zip](https://github.com/user-attachments/files/16428380/test.zip) file. Expected that `spaceBeforeSeparator=false` configuration option in `formatter-json.properties` is honoured.

Other configuration options work just fine.

**Versions (OS, Maven, Java, and others, as appropriate):**
 - Affected version(s) of this project: 2.24.1
 - OS: Windows 10 x64

**Additional information**

Additionally I have run the test [`JsonFormatterTest.testDoFormatFileWithConfig()`]() and at least in my setup it has generated the following JSON `target\testoutput\someFile.json` with spaces before field separator:
```
{
  "glossary" : {
    "GlossDiv" : {
      "GlossList" : {
        "GlossEntry" : {
          "Abbrev" : "ISO 8879:1986",
          "Acronym" : "SGML",
          "GlossDef" : {
            "GlossSeeAlso" : [
              "GML",
              "XML"
            ],
            "para" : "A meta-markup language, used to create markup languages such as DocBook."
          },
          "GlossSee" : "markup",
          "GlossTerm" : "Standard Generalized Markup Language",
          "ID" : "SGML",
          "SortAs" : "SGML"
        }
      },
      "title" : "S"
    },
    "title" : "example glossary"
  }
}
```
which is wrong because `spaceBeforeSeparator=false`:
![image](https://github.com/user-attachments/assets/ecca0f1a-0067-4c46-b03b-b29a33686334)

I have noticed that overridden function `DefaultPrettyPrinter#withSeparators(Separators separators)` is not called, which means that `_objectFieldValueSeparatorWithSpaces` is left with default Jackson value ` : `. This is because some of API has become deprecated in Jackson v2.17.
![image](https://github.com/user-attachments/assets/4b9cc1f4-195f-4b58-92c1-ef95c5c2dcea)

One of the solutions could be to initialize `_objectFieldValueSeparatorWithSpaces` directly in `createInstance()` as in [fix.patch](https://github.com/user-attachments/files/16428758/fix.patch).

ctubbsii:
@dmak Thanks for the detailed explanation. Would you be able to open a PR with your proposed patch for easier review and merging?

hazendaz:
fixed via #930

ctubbsii:
[...]

PR #930, opened 2024-08-07
Adapt field separator initialization to Jackson v2.17 API (#929)

dmak:
Fixes JSON formatting issue #929, namely, fixes field separator initialization as to Jackson v2.17 API.


hazendaz:
Thanks @dmak

PR #949, opened 2025-01-02
[maven] Update from component to inject per jsr330

hazendaz:
maven has deprecated component now.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,6 +129,11 @@
       <version>1.3.4</version>
     </dependency>
     <dependency>
+      <groupId>javax.inject</groupId>
+      <artifactId>javax.inject</artifactId>
+      <version>1</version>
+    </dependency>
+    <dependency>
       <groupId>net.revelc.code.formatter</groupId>
       <artifactId>jsdt-core</artifactId>
       <version>3.4.4</version>
--- a/src/main/java/net/revelc/code/formatter/FormatterMojo.java
+++ b/src/main/java/net/revelc/code/formatter/FormatterMojo.java
@@ -39,12 +39,13 @@ import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import javax.inject.Inject;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.maven.model.Resource;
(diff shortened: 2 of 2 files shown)


ctubbsii:
> maven has deprecated component now.

When you say "now", what version of Maven do you mean? It also seems very weird that Maven would deprecate Component in favor of javax.inject.Inject instead of jakarta.inject.Inject.

hazendaz:
From Google ai

Component annotation: This annotation was used to configure injection of Plexus components by MavenPluginManager. You can use JSR 330 annotations or Parameter with default expressions instead.

.....

Also directly deprecated https://maven.apache.org/plugin-tools/maven-plugin-annotations/apidocs/deprecated-list.html

Or

https://codehaus-plexus.github.io/plexus-containers/plexus-component-metadata/

Plexus itself is pretty much all deprecated.

Any maven 3.6.3+ if not earlier.  Majority of core maven plugins already had replaced it.

Since recent releases it's now flagged as such.

hazendaz:
Oh they are not on jakarta yet (namespace or othereise).  As far as the physical library they are still using the original javax copy from Google.  I reported that a long time ago.  They didn't care to upgrade the gav.  So, they are directly on the one with missing Metadata.  You can override that, but I think the configuration on this plugin makes exclusions and replacements more of a problem.  The one listed here is their physical usage.

hazendaz:
@ctubbsii FWIW maven fixed to use jakarta.inject 2.0.1 in maven 4.  Of course we will have to continue support for maven 3.  And there is a general understanding that plugins need to be separately supported.  That said, they also have javax.inject sitting in maven 4 as well.  So that goes to explain why they didn't use the correct one because it would be rather hard to do so.

pom.xml
```diff
@@ -128,6 +128,11 @@
       <artifactId>commons-logging</artifactId>
       <version>1.3.4</version>
     </dependency>
+    <dependency>
+      <groupId>javax.inject</groupId>
+      <artifactId>javax.inject</artifactId>
+      <version>1</version>
+    </dependency>
```
ctubbsii, line 135: It's unfortunate that javax didn't use a bom, like jakartaee does. I hate managing the javax dependency versions individually.

ctubbsii (approved):

PR #951, opened 2025-01-19
[pom] Update guava and commons beanutils

hazendaz:


hazendaz:
jdk 24/25 will be fixed by parent update which has just been done and will be updated separately.

PR #953, opened 2025-01-19
[pom] Bump revelc to parent to 7

hazendaz:


hazendaz:
ok groovy doesn't support java 25 yet so invoker fails there, but the parent update did allow support to java 24.

PR #960, opened 2025-01-20
Rewrite pom for better support along with updating outstanding library versions

hazendaz:
- Bump jdt core to 3.40.0
- Bump jsdt core to 3.5.0
- No longer exclude any libraries unless 100% necessary as tools do not show relationships and dependency convergence is on
- Do not worry about libraries we vetted that are ok with dependency convergence issues
- Drop part of the osgi overrides as no longer needed
- Use the last cglib to avoid any potential issues (at least until apache commons team finally releases digester upgrade after 14 years)  I expect this to be removed at that time.
- Override asm for same reason as cglib (the only required exclusion is here because group id change).  I expect this to be removed once digester new release is made.
- Adjust comments around how we handle convergence or override usage
- Drop commons lang3 as it was used once to check for empty, just check that directly

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -218,7 +218,7 @@
     <dependency>
       <groupId>org.eclipse.jdt</groupId>
       <artifactId>org.eclipse.jdt.core</artifactId>
-      <version>3.38.0</version>
+      <version>3.40.0</version>
     </dependency>
     <dependency>
       <groupId>org.jsoup</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -124,12 +124,6 @@
       <groupId>commons-beanutils</groupId>
       <artifactId>commons-beanutils</artifactId>
       <version>1.10.1</version>
-      <exclusions>
-        <exclusion>
-          <groupId>commons-logging</groupId>
-          <artifactId>commons-logging</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>commons-logging</groupId>
@@ -145,39 +139,11 @@
       <groupId>net.revelc.code.formatter</groupId>
       <artifactId>jsdt-core</artifactId>
       <version>3.5.0</version>
-      <exclusions>
-        <exclusion>
-          <groupId>com.google.errorprone</groupId>
-          <artifactId>error_prone_annotations</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>com.google.guava</groupId>
-          <artifactId>guava</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>org.osgi</groupId>
-          <artifactId>org.osgi.util.function</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>org.osgi</groupId>
-          <artifactId>osgi.annotation</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>org.slf4j</groupId>
-          <artifactId>slf4j-api</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>net.revelc.code.formatter</groupId>
       <artifactId>xml-formatter</artifactId>
       <version>0.4.0</version>
-      <exclusions>
-        <exclusion>
-          <groupId>org.slf4j</groupId>
-          <artifactId>slf4j-api</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>net.sourceforge.cssparser</groupId>
@@ -188,27 +154,11 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-digester3</artifactId>
       <version>3.2</version>
-      <exclusions>
-        <exclusion>
-          <groupId>commons-beanutils</groupId>
-          <artifactId>commons-beanutils</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>commons-logging</groupId>
-          <artifactId>commons-logging</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.codehaus.plexus</groupId>
       <artifactId>plexus-resources</artifactId>
       <version>1.3.0</version>
-      <exclusions>
-        <exclusion>
-          <groupId>org.codehaus.plexus</groupId>
-          <artifactId>plexus-utils</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.codehaus.plexus</groupId>
@@ -230,12 +180,6 @@
       <groupId>org.osgi</groupId>
       <artifactId>org.osgi.util.function</artifactId>
       <version>1.2.0</version>
-      <exclusions>
-        <exclusion>
-          <groupId>org.osgi</groupId>
-          <artifactId>osgi.annotation</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <!-- converge dependency with jsdt-core -->
--- a/pom.xml
+++ b/pom.xml
@@ -91,6 +91,12 @@
         <scope>import</scope>
       </dependency>
       <dependency>
+        <!-- override vulnerable transitive version from commons-digester3 -->
+        <groupId>commons-beanutils</groupId>
+        <artifactId>commons-beanutils</artifactId>
+        <version>1.10.1</version>
+      </dependency>
+      <dependency>
         <groupId>org.ow2.asm</groupId>
         <artifactId>asm</artifactId>
         <version>9.7.1</version>
@@ -130,12 +136,6 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
       <version>33.4.0-jre</version>
-    </dependency>
-    <dependency>
-      <!-- override vulnerable transitive version from commons-digester3 -->
-      <groupId>commons-beanutils</groupId>
-      <artifactId>commons-beanutils</artifactId>
-      <version>1.10.1</version>
     </dependency>
     <dependency>
       <groupId>commons-logging</groupId>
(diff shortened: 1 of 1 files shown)


hazendaz:
Will leave this out here for a week or two.  The pom is a lot smaller by going this route which allows using the IDE to look at dependency relationships (using Eclipse IDE).  The one that is not is asm due to the group id changing from the 14 year gap involved.

All libraries otherwise directly used were thoroughly vetted to see if any can be removed.  At this time no others can be removed.  There are a few that potentially can be with some additional work.  For example, plexus utils is used only to scan the directory tree which can be replaced by direct java code but I was not certain how to implement the walker along with the includes / exclusions.  Another library example, commons io is used one time to read a resource to a string, there is no NIO equivalent to pull that off in a one liner currently but tradeoff in that savings might not be worth it since less is better.  Guava is yet another case because java still doesn't support sha_512.  Another is plexus resources, used simply for reading resources that could be done internally.

The point of the above is to just express, a deep look into the usage was done to see if we can get rid of any other libraries and thus make this even faster than it is.  The remainder of this PR is simply to get it up to 2024-12 release and as such, this is ready for release immediately after this is merged.

pom.xml
```diff
@@ -98,6 +98,18 @@
...
+      <artifactId>cglib</artifactId>
+      <version>3.3.0</version>
+      <!-- This exclusion must remain as the group id changed. -->
+      <exclusions>
+        <exclusion>
+          <groupId>org.ow2.asm</groupId>
+          <artifactId>asm</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
```
ctubbsii, line 112: I don't know what cglib is, but it has some pretty serious warnings in its [source repo](https://github.com/cglib/cglib) that give me strong "we shouldn't be doing this" vibes:

> IMPORTANT NOTE: cglib is unmaintained and does not work well (or possibly at all?) in newer JDKs, particularly JDK17+. If you need to support newer JDKs, we will accept well-tested well-thought-out patches... but you'll probably have better luck migrating to something like [ByteBuddy](https://bytebuddy.net/).

pom.xml
```diff
@@ -273,7 +227,21 @@
             <phase>validate</phase>
             <configuration>
               <rules>
-                <dependencyConvergence />
+                <dependencyConvergence>
+                  <!-- The items listed below have been verified non issue for allowing. -->
```
ctubbsii, line 231: These might be a non-issue right now... but any changes to dependencies in the future could make them become an issue again... so this should not be done over the exclusions, which were done on a case-by-case basis.

pom.xml
```diff
@@ -283,14 +251,14 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-dependency-plugin</artifactId>
         <configuration>
+          <!-- The items listed below have been verified required to override. -->
```
ctubbsii, line 254: I don't know what this sentence means. The verb clause, "have been verified required to override" doesn't make grammatical sense to me.

pom.xml
```diff
@@ -283,14 +251,14 @@
...
           <ignoredUnusedDeclaredDependencies>
-            <!-- ignore transitive dependency overridden to avoid vulnerability -->
+            <ignored>cglib:cglib</ignored>
```
ctubbsii, line 256: What is cglib doing here? Why must it be ignored? Why is it added at all?

src/main/java/net/revelc/code/formatter/FormatterMojo.java
```diff
@@ -404,7 +403,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
         final var startClock = System.nanoTime();
 
-        if (StringUtils.isEmpty(this.encoding)) {
+        if (this.encoding == null || this.encoding.length() == 0) {
```
ctubbsii, line 406: :+1: to this change, if it avoids a whole dependency

hazendaz: will pull this one out and separately apply by itself.

pom.xml
hazendaz: cglib is already a transient dependency from commons digester but its not on any remotely modern copy.  So we are excluding the bad copy in favor of the last one released.

pom.xml
hazendaz: As noted in the comments to the overall pull request

"Use the last cglib to avoid any potential issues (at least until apache commons team finally releases digester upgrade after 14 years) I expect this to be removed at that time."

pom.xml
hazendaz: There is no difference in handling of exclusions vs case by case ignoring in this section.  The key its its in a single place not scattered throughout the pom.  That makes it actually easier to support.

pom.xml
hazendaz: for what commons-digester3 is doing now which is active today even, see https://github.com/apache/commons-digester/pull/9

pom.xml
```diff
@@ -116,12 +128,6 @@
       <groupId>commons-beanutils</groupId>
       <artifactId>commons-beanutils</artifactId>
       <version>1.10.0</version>
-      <exclusions>
```
hazendaz, line 119: we are already using commons-logging, so this exclusion was old and wrong.  As noted nothing actually checks for that, it requires humans to do so.

pom.xml
```diff
@@ -136,36 +142,12 @@
     <dependency>
       <groupId>net.revelc.code.formatter</groupId>
       <artifactId>jsdt-core</artifactId>
-      <version>3.4.4</version>
-      <exclusions>
```
hazendaz, line 140: Moved these to dependency convergence for same reason they were here.

pom.xml
hazendaz: see https://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html.  In our case, we already override everything, so this suffices.

pom.xml
hazendaz: Does this help understanding?  

          

or

hazendaz:
Please also review the comments as they have detailed reasons added to why each action was taken.  That becomes part of the history so others can see as to why it was done.

pom.xml
ctubbsii: These are not equivalent. The override controls the version pulled in by the resolver. The individual excludes ensure we have removed only those that are safely overridden by the version we choose. If any new ones show up in the dependency tree, we will need to verify that. If you suppress them in the rule, then we will not know that there are new occurrences to triage.

pom.xml
ctubbsii: Oh, I see. It's because you added cglib to the dependencies instead of the dependencyManagement section. I think it would need to be in the dependencyManagement section to manage the version for the resolver. It doesn't need to be ignored... since it never needs to be specified as a direct dependency.

pom.xml
ctubbsii: These are all equivalent, but does not help explain anything. More specific comments explaining why items are ignored would be useful. Some of our existing ignores have better comments. This doesn't add anything, because we can already presume they have been verified and are required to be here... or else they wouldn't be there. I think you can just omit the comment.

pom.xml
ctubbsii: I don't understand this comment. It wasn't being excluded because it wasn't being used. It was being excluded so we could specify a managed version, and it was excluded here so that the managed version did not conflict with this version in the dependencyConvergence rule.

Yes, it requires humans to check... but that's the point. A human should triage dependencyConvergence issues.

hazendaz:
@ctubbsii Rewrote to use mostly dependency management, removed the dependency plugin ignores entirely, left the ones we don't directly use or were not directly adherent within the eclipse parts.  This passes all the tests, is this more acceptable?

Note: The other PR open is entirely covered by this one.  I'd also say if good to go, release 3.40.0 here so we can move to 3.41.0 and address issues others noted in a couple areas.

hazendaz:
sorry didn't mean to close, clicked wrong button.

pom.xml
```diff
@@ -86,17 +86,62 @@
...
+        <artifactId>cglib</artifactId>
+        <version>3.3.0</version>
+      </dependency>
+      <dependency>
+        <!-- converge dependency with jsdt-core -->
+        <groupId>com.google.errorprone</groupId>
+        <artifactId>error_prone_annotations</artifactId>
+        <version>2.36.0</version>
+      </dependency>
+      <dependency>
```
hazendaz, line 118: Yes we use guava too but this feels save enough, its been years since guava made any breaking mistakes that mattered since really most of guava is in more recent java versions by default.

pom.xml
```diff
@@ -280,28 +238,19 @@
...
-            <ignore>commons-beanutils:commons-beanutils</ignore>
-            <!-- ignore alignment with commons-logging to support slf4j -->
-            <ignore>commons-logging:commons-logging</ignore>
-            <!-- ignore unused osgi items added only for transitive dependency convergence-->
-            <ignore>com.google.errorprone:error_prone_annotations</ignore>
-            <ignore>org.osgi:org.osgi.util.function</ignore>
-            <ignore>org.osgi:osgi.annotation</ignore>
-          </ignoredUnusedDeclaredDependencies>
-        </configuration>
-      </plugin>
```
hazendaz, line 304: note switching to dep management with this level of items did reduce the file by 51 lines.  So less to support.

ctubbsii (approved):
@hazendaz I added a commit to this PR that I think you'll find acceptable. I think it also addresses all my concerns as well.

hazendaz:
> @hazendaz I added a commit to this PR that I think you'll find acceptable. I think it also addresses all my concerns as well.

@ctubbsii Great!  Looks good.  Merging it now.  Do we want to wait a little while before cutting release or just go for it?  I'm available to do the release most nights this week.

ctubbsii:
> Do we want to wait a little while before cutting release or just go for it? I'm available to do the release most nights this week.

#907 is still open for this milestone. Is it fixed for this milestone?

hazendaz:
Hi @ctubbsii I believe that specific issue is resolved, probably in the last release you performed, but we probably should follow the instructions provided to confirm.  I can try to do so in a day or two but will ask original reporter if they can confirm its fixed in 2.25.0.

PR #977, opened 2025-04-27
[refactor] inline FormatterMojo::formatFile into FormatterMojo::execute to remove unnecessary indirection

ArcaneEcholan:
the FormatterMojo::formatFile was only called once and did not contain complex logic. inlining improves readability by eliminating an extra code jump.


ctubbsii (approved):
This looks fine, thanks.

PR #979, opened 2025-05-03
feat: support findup" cfg search strategy

ArcaneEcholan:


ctubbsii (commented):
I am opposed to these changes for the explanation described below:

> * existing strategies:
>   
>   * absolute config file path
>   * path config relative to project basedir

I would not characterize these as "strategies". The config file provided is the config file provided, regardless of whether it is absolute or relative. It's just the one file, exactly as the user specified. No searching, no discovery, just exactly what the user set.

> * new strategy:
>   
>   * **find-up** search: given a config filename (e.g. `java.xml`), search from `${project.basedir}` upward through parent directories
>   * search stops when a directory **without a `pom.xml`** is reached

So, first, this strategy is broken. It will stop as soon as a directory doesn't have a pom.xml. But it's not a requirement that every directory in a Maven multi-module project have a pom.xml. It's entirely possible to specify modules that are more than one layer deep, and for a child to refer to a parent that does not live above the child module. So, the whole premise of this strategy is flawed. It's also the case that just because a parent directory contains a pom.xml, it doesn't make it part of the same multi-module project. In fact, using git submodules would make this true much of the time, and a git submodule Maven project does not necessarily have any relation to the parent git project in which it exists. So, you really can't make any assumptions about the presence, or lack thereof, of any pom.xml files.

Second, this makes it much harder for users to determine which config file is being used, because it's not exactly what the user specified, but follows some auto-discovery strategy, making it a bit unpredictable. The current mechanism is to use the file that the user specified... no more, no less. It could not be simpler.

Third, this isn't necessary. There are so many plugins and techniques already available to find the root of a multi-module project, and to set it as a property, or to find the root of a git clone and set it as a property, etc. Those are much more predictable to use, since the file specified in the POM using those is much more clear to anybody troubleshooting, than any auto-discovery at-runtime determination of which config file to use. Adding a new feature to do auto-discovery adds complexity and makes it harder to actually figure out what's going on in the project, especially if this is relied upon in combination with one of the existing means to locate and configure config files.

While I appreciate the thoughtfulness and innovation put into this PR, I am strongly opposed to accepting this change for the above reasons. It's just not worth the added complexity or potential for error. We should keep the config location simple, and use exactly what the user specified, and nothing else. It's much easier for users to predict what's going on. If they want additional logic to auto-locate a config file, they can do that with a different plugin, and set a property, like what the [build-helper-maven-plugin rootLocation](https://www.mojohaus.org/build-helper-maven-plugin/rootlocation-mojo.html) goal does.

